PR #15, opened 2021-10-03
Migrate test database to same server as dev database

### LINKED ISSUE #2: Migrate test database to same server as dev database
## User Story and Context

As a developer, I want to run both the test and dev PostgreSQL databases on the same server (and within the same container).

Currently, we spin up a separate container called `db-test` to run our test database (which is used for integration testing). This unnecessarily uses extra system resources.

### Product
N/A

### Technical
* Both the dev and test databases run on the `sistering_db_1` host, with the dev DB named `sistering` and the test DB named `sistering_test`
* `db-test` is removed from `docker-compose.yml`
* All backend tests pass:
```bash
$ docker exec -it sistering_backend_1 /bin/bash -c "yarn test"
```

## Dev Notes
* Take a look at the accepted solution in this Stack Overflow [post](https://stackoverflow.com/questions/46668233/multiple-databases-in-docker-and-docker-compose). An init script is added to the image to create multiple databases.
* You will need to modify the DB connection string in `backend/testUtils/testDb.ts`

daniel-sm-yu:
## Migrate test database to same server as dev database
Closes #2

## Implementation description
* add script that creates multiple databases in one container
* in `docker-compose.yml`, remove `db-test` container and use script to add multiple databases to `db` container
* update `backend/testUtils/testDb.ts` and `backend/models/index.ts` with new database and host names from `.env`  

## Steps to test
1. run `docker-compose down --volumes` (to delete the stop the current containers and delete old databases)
2. run `docker-compose up` to run the updated `docker-compose.yml`
3.  run `docker exec -it sistering_backend_1 /bin/bash -c "yarn test"` to test the databases
       - in `testDb.ts` toggle between `process.env.POSTGRES_DB_TEST` and `process.env.POSTGRES_DB_DEV` to test both databases 

* `testDb` passing
* _Multiple database creation requested: sistering,sistering_test_ is outputted after `docker-compose up`

## Checklist


daniel-sm-yu:
test failing after resolving merge conflict (looking into it soon)

sherryhli:
@daniel-sm-yu Reproduced the error and I think I found the fix. The issue is that the `db-init/create-multiple-dbs.sh` script isn't being executed because it has CRLF line endings, can you try changing it it to LF?

In VS Code:
image

This may be a manual step that Windows users will have to do for the time being, but we can look into tweaking the `core.autocrlf` git configs

backend/models/index.ts
```diff
@@ -1,8 +1,7 @@
...
-);
-
\ No newline at end of file
+import * as path from "path";
+import { Sequelize } from "sequelize-typescript";
+
+export const sequelize = new Sequelize(
+  `postgres://${process.env.POSTGRES_USER}:${process.env.POSTGRES_PASSWORD}@${process.env.DB_HOST}:5432/${process.env.POSTGRES_DB_DEV}`,
+  { models: [path.join(__dirname, "/*.model.ts")] },
+);
```
sherryhli, line 7: Hmm diff looks a little weird, can you check if the committed file is using LF endings?

db-init/create-multiple-dbs.sh
```diff
@@ -0,0 +1,22 @@
...
+EOSQL
+}
+
+if [ -n "$POSTGRES_MULTIPLE_DATABASES" ]; then
+	echo "Multiple database creation requested: $POSTGRES_MULTIPLE_DATABASES"
+	for db in $(echo $POSTGRES_MULTIPLE_DATABASES | tr ',' ' '); do
+		create_user_and_database $db
+	done
+	echo "Multiple databases created"
+fi
```
sherryhli, line 22: Nit: last line of file should always be a newline

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: "3.7"
+version: '3.7'
```
sherryhli, line 1: Keep as double quotes please

sherryhli (commented):
A couple of minor things

backend/models/index.ts
daniel-sm-yu: image
For sure, looks like it is LF

docker-compose.yml
daniel-sm-yu: updated my prettier to use double quotes

daniel-sm-yu:
> @daniel-sm-yu Reproduced the error and I think I found the fix. The issue is that the `db-init/create-multiple-dbs.sh` script isn't being executed because it has CRLF line endings, can you try changing it it to LF?
> 
> In VS Code: image
> 
> This may be a manual step that Windows users will have to do for the time being, but we can look into tweaking the `core.autocrlf` git configs

Nice! Would've never thought it would be because of line endings - thanks!!!

sherryhli (approved):
LGTM, 🚢 it!

PR #36, opened 2021-10-17
Prisma user and entity models

### LINKED ISSUE #14: Migrate user and entity models and services to use Prisma
## User Story and Context

Following #8, we have decided to migrate from Sequelize to Prisma for our ORM. #12 installed Prisma as a dependency and tested the connection logic, now we need to migrate existing models and services to use Prisma.

### Product
N/A

### Technical
* `User` and `Entity` models are defined using Prisma instead of Sequelize
* `UserService` and `EntityService` use Prisma client methods instead of Sequelize client methods for CRUD operations
* Public APIs over users and entities remain unchanged (test all queries and mutations [here](https://uwblueprint.github.io/starter-code-v2/docs/user-auth/testing.html#graphql) and [here](https://uwblueprint.github.io/starter-code-v2/docs/crud/testing.html#graphql) to ensure their behaviours have not changed)
* All references to Sequelize methods are removed from the code

## Dev Notes

You'll largely be following this official Prisma [guide](https://www.prisma.io/docs/getting-started/setup-prisma/add-to-existing-project/relational-databases-typescript-postgres). Please read it through **and the additional notes below** before starting this task.

While developing, you should be running the application with Docker and running all yarn and npx commands from the backend container:
```
# from the root of the sistering folder
$ docker-compose up --build

# get the backend container name, it should be sistering_backend_1
$ docker ps

# run a bash shell in the backend container
$ docker exec -it <CONTAINER-NAME> /bin/bash
```

Additional Notes:
* Wait for #13 to be merged into main before branching off
* You'll need to add the following line to `backend/prisma/.env`
```
DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${DB_HOST}:5432/${POSTGRES_DB}"
```
* To create the Prisma models, you can use [introspection](https://www.prisma.io/docs/getting-started/setup-prisma/add-to-existing-project/relational-databases/introspection-typescript-postgres), however, please check the following for the generated models:
    * Sequelize automatically included `createdAt` and `updatedAt` columns for all tables but we don't currently have any use for them so they can be excluded from the Prisma model
    * Following TypeScript/JavaScript convention, we use camelCase field names in the model, but following PostgreSQL convention, we use snake_case for database columns. Similarly, while the model name starts with a capital letter according to convention, the table name should be all lowercase. Ensure that these differences are reflected in the model, you may need to use the `@map` directive, [example](https://github.com/prisma/prisma/discussions/2530):
```
model Post {
  id       Int      @default(autoincrement()) @id
  authorId Int?     @map("author_id")
  isActive Boolean? @default(false) @map("is_active")

  @@map("posts")
}
```
* At this point, prior to running Prisma migrations, you will likely want to comment out or delete the Sequelize initialization logic in server.ts. You'll also want to take down your database and all of its current data with `docker-compose down --volumes`
* Running `docker-compose down --volumes` will delete your test admin user, so to test the GraphQL API, create a new user for yourself by running this insert statement in the DB container
```sql
$ docker exec -it sistering_db_1 /bin/bash -c "psql -U postgres -d sistering"

INSERT INTO users (first_name, last_name, auth_id, role) VALUES ('<FIRST_NAME>', '<LAST_NAME>', '<FIND-YOUR-UID-FROM-FIREBASE>', 'Admin');
```

lambo-liu:
## Ticket link

Closes #14 

## Implementation description
* Generated Prisma schemas for `user` and `entity`
* Migrated `userService` from Sequelize to Prisma
* Migrated `entityService` from Sequelize to Prisma

## Steps to test
1. Go to http://localhost:5000/graphql and check that the `user` and `entity` CRUD services are still functional (any of login, users, userByEmail, createUser, updateUser, deleteUser, entities, entity, createEntity, updateEntity, deleteEntity)

* Correct usage of Prisma operations, correct Prisma schemas

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR


backend/prisma/schema.prisma
```diff
@@ -1,11 +1,42 @@
...
 
 datasource db {
   provider = "postgresql"
   url      = env("DATABASE_URL")
 }
 
-generator client {
-  provider = "prisma-client-js"
+model Entity {
+  id               Int      @id @default(autoincrement())
```
tamim-khan, line 11:
```suggestion
  id               String      @id @default(uuid())
```
Might want to use UUIDs for Ids instead of a serial id.
@sherryhli any thoughts?

backend/prisma/schema.prisma
```diff
@@ -1,11 +1,42 @@
...
+  enumField        Letters  @map("enum_field")
+  stringArrayField String[] @map("string_array_field")
+  boolField        Boolean  @map("bool_field")
+  fileName         String   @map("file_name")
+
+  @@map("entities")
+}
+
+model User {
+  id        Int    @id @default(autoincrement())
```
tamim-khan, line 23: Same here

backend/services/implementations/entityService.ts
```diff
@@ -125,49 +131,50 @@ class EntityService implements IEntityService {
...
-          file_name: fileName,
+      updateResult = await prisma.entity.update({
+        where: { id: Number(id) },
+        data: {
+          stringField: entity.stringField,
+          intField: entity.intField,
+          enumField: entity.enumField,
+          stringArrayField: entity.stringArrayField,
+          boolField: entity.boolField,
+          fileName,
```
sherryhli, line 142: May want to put the `currentEntity` lookup and update operation in a transaction, seems like a good use case for Prisma's interactive transactions feature:
https://www.prisma.io/docs/concepts/components/prisma-client/transactions#interactive-transactions-in-preview

I'll create a new ticket for this.

backend/services/implementations/entityService.ts
```diff
@@ -125,49 +131,50 @@ class EntityService implements IEntityService {
...
   async deleteEntity(id: string): Promise<void> {
     try {
-      const entityToDelete = await PgEntity.findByPk(id, { raw: true });
-      const deleteResult: number | null = await PgEntity.destroy({
-        where: { id },
+      const entityToDelete = await prisma.entity.findUnique({
+        where: { id: Number(id) },
+      });
+      const deleteResult: Entity | null = await prisma.entity.delete({
+        where: { id: Number(id) },
```
sherryhli, line 170: Prisma's `delete` method will return the deleted object so I don't think we need the separate query for `entityToDelete` now.

backend/services/implementations/userService.ts
```diff
@@ -1,8 +1,11 @@
 import * as firebaseAdmin from "firebase-admin";
+import { PrismaClient, User } from "@prisma/client";
 import IUserService from "../interfaces/userService";
 import { CreateUserDTO, Role, UpdateUserDTO, UserDTO } from "../../types";
 import logger from "../../utilities/logger";
-import User from "../../models/user.model";
+// import User from "../../models/user.model";
```
sherryhli, line 6: Delete

sherryhli (commented):
Left a couple of comments but nothing that needs to be urgently addressed. Good call to merge and unblock.

backend/server.ts
```diff
@@ -36,13 +38,25 @@ server.applyMiddleware({
...
+main()
+  .catch((e) => {
+    // e should be a PrismaClientInitializationError if connection fails
+    console.error("Error connecting to DB using Prisma!");
+    throw e;
+  })
+  .finally(async () => {
+    await prisma.$disconnect();
+  });
+
```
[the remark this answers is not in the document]
daniel-sm-yu: should I also remove
```
import { PrismaClient } from "@prisma/client";

const prisma = new PrismaClient();
```
these lines become unused without the code on lines 46 and 57

sherryhli:
@daniel-sm-yu would you mind also deleting the unused Sequelize model files?

backend/prisma/schema.prisma
tamim-khan: I think serial Ids are generally a bad pattern

PR #55, opened 2021-11-27
Create Shift CRUD Methods and GraphQL Resolvers

### LINKED ISSUE #51: Create CRUD service and GraphQL resolvers for Shifts
## User Story and Context

We want to write application code for creating/reading/updating/deleting shifts (#42).

### Product
See #42.

### Technical
* Create `ShiftsService` which contains CRUD methods acting on the `shifts` table
    * Bulk create `create(startTime, endTime, startDate, endDate, reccurrenceInterval)`
    * `getByPostingId` (return all shifts for a posting), `getById` (return single shift by shift Id)
    * `updateById`, `updateByPostingId` (delete all existing shfits, then recreate in same manner as `create` method above)
    * `deleteById`, `deleteByPostingId`
* Define `ShiftRequestDTO` and `ShiftResponseDTO` types
* Create GraphQL resolvers for queries and mutations that call into the  `ShiftsService` CRUD methods. These GraphQL operations should be restricted to admins.

## Dev Notes
* N/A

lambo-liu:
## Ticket link

Closes #51 

## Implementation description
* Implemented the following endpoints & CRUD methods for shifts:
> getShifts
> getShift
> createShifts (creates shifts of a posting by bulk)
> updateShift (updates a single shift)
> updateShifts (updates shifts of a posting by bulk)
> deleteShift (deletes a single shift)
> deleteShifts (deletes all shifts of a posting)

## Steps to test
1. Create a dummy posting using SQL. Here are the following SQL queries to run:
`
insert into branches (id, name) values (1, 'test branch');
`
`
insert into postings (id, branch_id, type, title, description, auto_closing_date, start_date, end_date) values (1, 1, 'GROUP', 'a', 'a', '2022-01-09T21:33:51.302Z', '2022-01-09T21:33:51.302Z', '2022-01-09T21:33:51.302Z');
`
2. Go to `localhost:5000/graphql`, and test the individual queries & resolvers.

Examples:
Create shifts by bulk, using a biweekly recurrence interval
`mutation {
  createShifts(shifts: {
    postingId: "1",
    times: [
      {
        date: "2021-12-16"
        startTime: "01:30"
        endTime: "04:30"
      },
      {
        date: "2021-12-18"
        startTime: "12:00"
        endTime: "16:00"
      }
    ],
    endDate: "2022-01-15",
    recurrenceInterval: BIWEEKLY
  }) {
    id
    postingId
    startTime
    endTime
  }
}`

Update a single shift
`mutation {
  updateShift(id: "1", shift: {
    startTime: "2021-12-16 04:30",
    endTime: "2021-12-16 06:30"
  }) {
    postingId
    startTime
    endTime
  }
}`

* There are lots of validations to be checked. For bulk create / update, make sure that users can only input time blocks within a 1-week timeframe.
* If duplicate DateTimes persist, they should be neglected in the bulk creation process.
* Create/Update procedures should only work with a valid postingId
* Times should be inputted in the `HH:mm` format, and DateTimes should be inputted in the `YYYY-MM-DD HH:mm` format. Any other format should throw an error.
* Check that the functionality checks out and the code is clean!
* Have a good day 🌈 

## Checklist


backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+  ): Promise<ShiftResponseDTO[] | null> {
+    // Validate postingId
+    const posting = await prisma.posting.findUnique({
+      where: {
+        id: Number(shifts.postingId),
+      },
+    });
+    if (!posting) {
+      throw new Error(`Posting ${shifts.postingId} not found.`);
+    }
```
sherryhli, line 290: Not needed due to foreign key constraint :)

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    if (!this.validateTimeBlocks(shifts.times))
+      throw new Error("Invalid time blocks");
+
+    await this.deleteShifts(postingId);
+    const newShifts: ShiftResponseDTO[] = await this.createShifts(shifts);
```
sherryhli, line 296: Instead of calling the `deleteShifts` and `createShifts` methods, I think we should call the underlying Prisma methods directly within a transaction to ensure that the update is treated as a single atomic operation, leaving the database with an always consistent state

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    return newShifts;
+  }
+
+  async deleteShift(shiftId: string): Promise<void> {
+    try {
+      const shiftToDelete = await prisma.shift.findUnique({
+        where: { id: Number(shiftId) },
+      });
```
sherryhli, line 305: This query isn't needed

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+      const deleteResult: Shift | null = await prisma.shift.delete({
+        where: { id: Number(shiftId) },
+      });
+
+      if (!shiftToDelete || !deleteResult) {
+        throw new Error(`Shift id ${shiftId} not found`);
+      }
```
sherryhli, line 312: Exception will be thrown if shift ID is not found, catch block already handles this

Ref: https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#return-type-6

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+    try {
+      // Validate postingId
+      const posting = await prisma.posting.findUnique({
+        where: {
+          id: Number(postingId),
+        },
+      });
+      if (!posting) {
+        throw new Error(`Posting ${postingId} not found.`);
+      }
```
sherryhli, line 329: See previous comments :)

backend/services/interfaces/IShiftService.ts
```diff
@@ -0,0 +1,106 @@
...
+  unit: DurationInputArg2;
+};
+
+interface IShiftService {
+  /**
+   * Get moment.js method parameters using recurrence interval
+   * @param recurrence a recurrence interval
+   * @returns value & unit of time
+   */
+  getDuration(recurrence: RecurrenceInterval): DurationArgs;
```
sherryhli, line 21: Minor stylistic preference but I don't think we should expose helper functions in the interface, which acts as a guarantee of what methods are publicly available from `ShiftService`.

sherryhli (changes requested):
Great progress so far! 😄 This is a big set of changes so I'll be doing another round of review after these comments are addressed

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,348 @@
...
+    )
+      return false;
+
+    // Check that start dates are within the range of 1 week
+    let earliestDate = moment(timeBlocks[0].date, "YYYY-MM-DD", true);
+    let latestDate = moment(timeBlocks[0].date, "YYYY-MM-DD", true);
+
+    if (!earliestDate.isValid() || !latestDate.isValid()) {
+      return false;
+    }
```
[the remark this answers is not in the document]
lambo-liu: this should be checked in the `timeBlocks.some` block right above this

lambo-liu:
@sherryhli updated changes! Feel free to review again. Thanks! 💯

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,347 @@
...
+
+  async updateShifts(
+    postingId: string,
+    shifts: ShiftBulkRequestDTO,
+  ): Promise<ShiftResponseDTO[] | null> {
+    const [valid, errorMessage] = this.validateTimeBlocks(shifts.times);
+    if (!valid) throw new Error(errorMessage);
+
+    return prisma.$transaction(async (prismaClient) => {
+      await prismaClient.shift.deleteMany({
```
sherryhli, line 289: I took a closer look at the [docs](https://www.prisma.io/docs/guides/performance-and-optimization/prisma-client-transactions-guide) and it looks like we don't need to directly call Prisma methods within the transaction block, calling other functions work too. We can simplify to calling `this.deleteShifts` and `this.createShifts` here.

backend/services/interfaces/IShiftService.ts
```diff
@@ -0,0 +1,90 @@
+import { DurationInputArg1, DurationInputArg2 } from "moment";
+import {
+  RecurrenceInterval,
+  ShiftBulkRequestDTO,
+  ShiftRequestDTO,
+  ShiftResponseDTO,
+  TimeBlockDTO,
+} from "../../types";
+
+export type DurationArgs = {
```
sherryhli, line 10: Can this be defined locally in the `implementations/shiftService.ts` file?

sherryhli (commented):
One more minor comment!

backend/graphql/resolvers/shiftResolvers.ts
```diff
@@ -0,0 +1,58 @@
...
+    ): Promise<void> => {
+      return shiftService.deleteShift(id);
+    },
+    deleteShifts: async (
+      _parent: undefined,
+      { id }: { id: string },
+    ): Promise<void> => {
+      return shiftService.deleteShifts(id);
+    },
+  },
```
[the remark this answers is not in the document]
sherryhli: GraphQL best practice is to always return a value from operations so for deletions, let's return the ID of the deleted item. Unfortunately starter code did not follow this convention 😅

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,347 @@
...
+          !tb.date ||
+          !tb.startTime ||
+          !tb.endTime ||
+          !moment(
+            `${tb.date} ${tb.startTime}`,
+            dateTimeFormat,
+            true,
+          ).isValid() ||
+          !moment(`${tb.date} ${tb.endTime}`, dateTimeFormat, true).isValid() ||
+          moment(tb.startTime, timeFormat) >= moment(tb.endTime, timeFormat),
```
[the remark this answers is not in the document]
lambo-liu: I don't think so? `startTime` and `endTime` are both in timeFormat and in the same day `date`. So I think when they're parsed they will always be in the same day.

lambo-liu:
@sherryhli fixed! Let me know if there's anything else 👍

backend/yarn.lock
```diff
@@ -4741,6 +4741,119 @@ mkdirp@^0.5.1:
...
+  version "0.5.33"
+  resolved "https://registry.yarnpkg.com/moment-timezone/-/moment-timezone-0.5.33.tgz#b252fd6bb57f341c9b59a5ab61a8e51a73bbd22c"
+  integrity sha512-PTc2vcT8K9J5/9rDEPe5czSIKgLoGsH8UNpA4qZTVw0Vd/Uz19geE9abbIOQKaAQFcnQ3v5YEXrbSc5BpshH+w==
+  dependencies:
+    moment ">= 2.9.0"
+
+"moment@>= 2.9.0", moment@^2.26.0, moment@^2.29.1:
+  version "2.29.1"
+  resolved "https://registry.yarnpkg.com/moment/-/moment-2.29.1.tgz#b2be769fa31940be9eeea6469c075e35006fa3d3"
+  integrity sha512-kHmoybcPV8Sqy59DwNDY3Jefr64lK/by/da0ViFcuA4DH0vQg5Q6Ze5VimxkfQNSC+Mls/Kx53s7TjP1RhFEDQ==
```
sherryhli, line 4759: This is the only yarn.lock dependency change needed, the rest are unused and were deleted in #56

backend/Dockerfile
```diff
@@ -6,6 +6,8 @@ COPY package.json yarn.lock tsconfig.json ./
 
 RUN apt-get update && apt-get install -y libssl-dev
 
+RUN apt-get update && apt-get install -y libcurl3
```
sherryhli, line 9: Let's remove the `libssl-dev` dependency above, it should be pulled transitively by the `libcurl3` install

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,372 @@
...
+    // Get moment's duration function args using recurrence
+    const duration: DurationArgs | null = this.getDuration(
+      shifts.recurrenceInterval,
+    );
+
+    const shiftTimes: TimeBlock[] = shifts.times.flatMap((time) => {
+      const startTime = moment(
+        `${time.date} ${time.startTime}`,
+        dateTimeFormat,
+        true,
```
sherryhli, line 130:
```suggestion
        STRICT_MODE,
```

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,372 @@
...
+      );
+      const endTime = moment(
+        `${time.date} ${time.endTime}`,
+        dateTimeFormat,
+        true,
```
sherryhli, line 135:
```suggestion
        STRICT_MODE,
```

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (
+      timeBlocks.some(
+        (tb) =>
+          !tb.date ||
+          !tb.startTime ||
+          !tb.endTime ||
+          !moment(
+            `${tb.date} ${tb.startTime}`,
+            dateTimeFormat,
+            true,
```
sherryhli, line 75:
```suggestion
            STRICT_MODE,
```

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,372 @@
...
+          let start = startTime.clone();
+          start < endDate;
+          start.add(duration.value, duration.unit)
+        ) {
+          recurringShifts.push({
+            startTime: start.toDate(),
+            endTime: end.toDate(),
+          });
+          end = end.add(duration.value, duration.unit);
+        }
```
sherryhli, line 150: Can declare and increment both start and end in loop initialization. Side note: is the `startTime.clone()` necessary?

```suggestion
        for (
          let start = startTime.clone(), end = endTime;
          start < endDate;
          start.add(duration.value, duration.unit), end.add(duration.value, duration.unit)
        ) {
          recurringShifts.push({
            startTime: start.toDate(),
            endTime: end.toDate(),
          });
        }
```

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,372 @@
...
+
+    return prisma.$transaction(async (prismaClient) => {
+      await prismaClient.shift.deleteMany({
+        where: { postingId: Number(postingId) },
+      });
+
+      const newShifts = await Promise.all(
+        shiftTimes.map(async (shiftTime: TimeBlock) => {
+          try {
+            const newShift = await prismaClient.shift.create({
```
sherryhli, line 321: I think we still need to check that the shift created here is not a duplicate of an existing shift (copy validation logic in `createShift`)

sherryhli (commented):
Sorry for the delay reviewing! A few very minor things left, which I can take of

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,402 @@
...
+          startTime: startTime.toDate(),
+          endTime: endTime.toDate(),
+        });
+      }
+      return recurringShifts;
+    });
+    return shiftTimes;
+  }
+
+  async checkConflictingShifts(
```
sherryhli, line 172: @LenaNguyen some context on why we check for conflicts :)

We want to avoid creating duplicate shifts (i.e. 2 shifts associated with the same posting that have the exact same start time and end time. This validation is pretty costly but unless there are significant performance issues, I think it's worthwhile to keep. We wouldn't want the database to be spammed with lots of duplicate entries (either accidentally or maliciously), it'd hurt performance and waste our row count quota in Heroku.

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,402 @@
...
+    );
+
+    if (moment.duration(latestDate.diff(earliestDate)).asDays() >= 7)
+      return [false, "Time blocks must be within a week"];
+
+    return [true, ""];
+  }
+
+  buildTimeBlocks(shifts: ShiftBulkRequestDTO): TimeBlock[] {
+    const endDate: Moment = moment(shifts.endDate, dateFormat, STRICT_MODE).add(
```
LenaNguyen, line 127: why do we add a day to the endDate?

backend/services/implementations/shiftService.ts
```diff
@@ -0,0 +1,402 @@
...
+      );
+      const endTime = moment(
+        `${time.date} ${time.endTime}`,
+        dateTimeFormat,
+        STRICT_MODE,
+      );
+      const recurringShifts = [];
+      if (duration) {
+        for (
+          let start = startTime.clone(), end = endTime;
```
LenaNguyen, line 151: why do we copy startTime but not endTime?

LenaNguyen (commented):
> Create a dummy posting using SQL. Here are the following SQL queries to run:
insert into branches (id, name) values (1, 'test branch');
insert into postings (id, branch_id, type, title, purpose, responsibilitie, tasks, desired_qualities, mandatory_activities, benefits, working_conditions, num_volunteers) values (1, 1, 'GROUP', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 0);
Go to localhost:5000/graphql, and test the individual queries & resolvers.

The insert for postings in this PR description is out of date right? I tried running it and the table no longer has some of these colums.

lambo-liu:
> > Create a dummy posting using SQL. Here are the following SQL queries to run:
> > insert into branches (id, name) values (1, 'test branch');
> > insert into postings (id, branch_id, type, title, purpose, responsibilitie, tasks, desired_qualities, mandatory_activities, benefits, working_conditions, num_volunteers) values (1, 1, 'GROUP', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 'a', 0);
> > Go to localhost:5000/graphql, and test the individual queries & resolvers.
> 
> The insert for postings in this PR description is out of date right? I tried running it and the table no longer has some of these colums.

Sorry Lena! Forgot to update these after Brian & Albert changed the schema for postings. Will update this now.

backend/services/implementations/shiftService.ts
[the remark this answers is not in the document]
lambo-liu: This offset is made so that when we compare `startTime` to `endDate`, any time block that starts and ends on the original endDate will be added to `shiftTimes`. For example, if our end date is 2022-01-01, a time block that starts on that same day (2022-01-01) will be added as it passes the `start < endDate` check.

backend/services/implementations/shiftService.ts
lambo-liu: We're setting `start` to the value of `startTime` (hence cloning `startTime`) here because we're using `start` as an iterator to the for loop.

backend/services/implementations/shiftService.ts
lambo-liu: I asked product peops last month and they were OK with overlapping shift times 👍

sherryhli: > Not sure what is more efficient but we can also add an index to the database so that postingId, startTime, and endTime are unique.

I like this idea. My feeling is that even with the write overhead of indexing, it might be more performant as it wouldn't require querying all shifts of a posting and then comparing start/end times for every shift write.

Would suggest merging this current implementation first though and doing some experiments before committing to the index approach.

backend/services/implementations/shiftService.ts
sherryhli: I think it's safe to not clone `startTime` but I wasn't 100% sure, @lambo-liu can you confirm?

sherryhli:
> The insert for postings in this PR description is out of date right? I tried running it and the table no longer has some of these colums.

@LenaNguyen this is the updated insert:
```sql
INSERT INTO postings (branch_id, type, title, description, num_volunteers, auto_closing_date, start_date, end_date) VALUES (1, 'GROUP', 'Title', 'Description', 5, '2021-01-01', '2021-01-11', '2021-01-15');
```

Could you take a final pass at this PR, I think it's good to go :)

sherryhli (approved):
LGTM but let's hold off on merging until Lena approves

LenaNguyen (approved):
Looks good :))

PR #90, opened 2022-01-16
Built out shift signup endpoints

### LINKED ISSUE #54: Create SignupService and corresponding GraphQL resolvers
## User Story and Context

Given a shift, we want to allow volunteers to signup, and allow both volunteers and employees to edit properties such as signup status.

### Product
N/A

### Technical
* Add a `note`  (varchar) column to the `Signup` Prisma model
* There are the following new DTO types
```
type CreateShiftSignupRequestDTO = {
  shiftId
  userId
  numVolunteers
  note
}

type UpdateShiftSignupRequestDTO = {
  numVolunteers
  note
  status
}

type ShiftSignupResponseDTO = {
  shiftId
  userId
  numVolunteers
  note
  status
}
```
* There are service methods for creating, updating, and reading signups
```
// bulk create signups is "all of nothing" operation, all signups must be successfully persisted for this operation to succeed
createShiftSignups(Array<CreateShiftSignupRequestDTO>) : Array<ShiftSignupResponseDTO >
updateShiftSignup(shiftId, userId, UpdateShiftSignupRequestDTO): ShiftSignupResponseDTO 
getShiftSignupsForUser(userId): Array<ShiftSignupResponseDTO >
```
* There are corresponding GraphQL queries and mutations calling into the new methods

## Dev Notes
* N/A

hujoseph99:
## Ticket link

Closes #54

## Implementation description
* Defined `CreateShiftSignupRequestDTO`, `UpdateShiftSignupRequestDTO`, `ShiftSignupResponseDTO` types
* Implemented the following CRUD endpoints with their corresponding graphql queries and mutations:
  * `createShiftSignups`
  * `updateShiftSignup`
  * `getShiftSignupsForUser` 
* Prisma schema changes (with corresponding migrations):
  * Added `note` varchar to `Signup` prisma model
  * Added foreign key constraint to `userId` in `Signup` model.
  * Added a unique key constraint to for `shiftId` and `userId` pair in `Signup` model

## Steps to test
1. Ensure that the docker containers are up and running
2. Run prisma `migration` and `generate` commands
2. Ensure that you have test data in the following tables: `users`, `shifts`, `postings`, `branches`.
  - If you don't have data, these commands might be helpful for getting some in easily:
  - Through the Postgres CLI:
    - `insert into branches (id, name) values (1, 'test branch');`
    - `insert into postings (id, branch_id, type, title, description, auto_closing_date, start_date, end_date, status) values (1, 1, 'GROUP', 'a', 'a', '2022-01-09T21:33:51.302Z', '2022-01-09T21:33:51.302Z', '2022-01-09T21:33:51.302Z', 'DRAFT');`
  - Through graphql interface (`localhost:5000/graphql`):
    - `mutation {
  createShifts(
    shifts: {
      postingId: "1"
      times: [
        { startTime: "2022-01-25T13:30", endTime: "2022-01-25T15:00" }
        { startTime: "2022-01-26T12:00", endTime: "2022-01-26T16:00" }
      ]
      endDate: "2022-03-15"
      recurrenceInterval: BIWEEKLY
    }
  ) {
    id
    postingId
    startTime
    endTime
  }
}
`
  - Credits to @lambo-liu in #55 
4. Go to `localhost:5000/graphql` and test the following new endpoints: `createShiftSignups`, `updateShiftSignup`, and `getShiftSignupsForUser`.  The following are some query/mutations that I wrote in my own testing that you might be able to use:
```
query {
  getShiftSignupsForUser(userId: "1") {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}

mutation {
  createShiftSignups(
    shifts: [{ shiftId: 2, userId: 1, numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}

mutation {
  updateShiftSignup(
    shiftId: "2"
    userId: "1"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
```

* Ensure that it works for edge cases like duplicating shift signups, signing up for a shift that doesn't exist, a user that doesn't exist signing up for a shift, etc.
* General testing for basic features

## Checklist


backend/services/implementations/shiftSignupService.ts
```diff
@@ -0,0 +1,88 @@
...
+        where: { userId: Number(userId) },
+      });
+      return shiftSignups.map((signup) => this.convertSignupToDTO(signup));
+    } catch (error) {
+      Logger.error(`Failed to shift signups. Reason = ${error.message}`);
+      throw error;
+    }
+  }
+
+  async createShiftSignups(
```
LenaNguyen, line 40: @sherryhli @JoshuaYe are we allowing non-volunteer roles to be added to shifts as well? Otherwise we should add a check for their role.

LenaNguyen (dismissed):
Just wait for this comment to be addressed. Everything looks great!

backend/services/implementations/shiftSignupService.ts
sherryhli: To my knowledge, non-volunteers should **not** be able to sign up, so checking would be a good idea. I wonder if we can perform the check in the authorization layer instead though (assuming we don't have a case where an admin is able to sign up on behalf of a volunteer). This saves a lookup query to find the role of the user.

backend/prisma/migrations/20220121012944_add_volunteer_users/migration.sql
```diff
@@ -98,9 +98,6 @@ CREATE UNIQUE INDEX "employees_id_key" ON "employees"("id");
 -- CreateIndex
 CREATE UNIQUE INDEX "volunteers_id_key" ON "volunteers"("id");
 
--- AddForeignKey
-ALTER TABLE "signups" ADD CONSTRAINT "signups_userId_fkey" FOREIGN KEY ("userId") REFERENCES "users"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
-
```
sherryhli, line 103: Was this deletion autogenerated? We shouldn't ever be modifying existing migration files as it will break our deployed environments.

backend/prisma/migrations/20220116032551_add_note_to_signup/migration.sql
```diff
@@ -0,0 +1,8 @@
+/*
```
sherryhli, line 1: Can you please update the timestamp in the migration folder name to be after `20220121012944`, which is the last migration to run in the staging environment? Otherwise the 2 signup table migrations will not run at all.

backend/prisma/migrations/20220116173256_added_unique_and_forign_key_contraints_to_signup/migration.sql
```diff
@@ -0,0 +1,11 @@
...
+  Warnings:
+
+  - A unique constraint covering the columns `[shifts_id,userId]` on the table `signups` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- CreateIndex
+CREATE UNIQUE INDEX "signups_shifts_id_userId_key" ON "signups"("shifts_id", "userId");
+
+-- AddForeignKey
+ALTER TABLE "signups" ADD CONSTRAINT "signups_userId_fkey" FOREIGN KEY ("userId") REFERENCES "users"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
sherryhli, line 11: Please update the timestamp of this migration's folder as well. After doing that, this line here should no longer be necessary as the FK would have been added as part of `backend/prisma/migrations/20220121012944_add_volunteer_users/migration.sql`

backend/prisma/schema.prisma
```diff
@@ -131,8 +131,10 @@ model Signup {
   userId        Int
   status        SignupStatus
   numVolunteers Int          @default(1) @map("num_volunteers")
+  note          String
 
   @@map("signups")
+  @@unique([shiftId, userId])
```
sherryhli, line 137: Thoughts about making `[shiftId, userId]` the PK of this table?

sherryhli (changes requested):
Just jumping in to point out a small issue with the order of migrations 🙂

backend/services/implementations/shiftSignupService.ts
JoshuaYe: Sherry is correct. If a volunteer has not requested a shift, they cannot be assigned to it by an admin. This was confirmed by Aoife in our last client meeting.

sherryhli: Got it, thanks @JoshuaYe! In that case I think handling it with the authorization middleware would be perfect, we check that the `CreateShiftSignups` mutation can only be invoked by a user with the VOLUNTEER role, and that a user can only sign themselves up (i.e. userId in request body === userId corresponding to access token)

hujoseph99: Updated

backend/prisma/migrations/20220121012944_add_volunteer_users/migration.sql
hujoseph99: Sorry, this was my misunderstanding with migrations. I've fixed it now

backend/prisma/migrations/20220116032551_add_note_to_signup/migration.sql
hujoseph99: Updated

backend/prisma/migrations/20220116173256_added_unique_and_forign_key_contraints_to_signup/migration.sql
hujoseph99: Updated

backend/prisma/schema.prisma
hujoseph99: I think it makes sense, I've made the changes

hujoseph99:
Sorry for the late response, I've been busy these past two days. I've made the changes now!

backend/graphql/index.ts
```diff
@@ -128,6 +132,7 @@ const authorizedByAllRoles = () =>
 const authorizedByAdmin = () => isAuthorizedByRole(new Set(["ADMIN"]));
 const authorizedByAdminAndVolunteer = () =>
   isAuthorizedByRole(new Set(["ADMIN", "VOLUNTEER"]));
+const authorizedByVolunteer = () => isAuthorizedByRole(new Set(["VOLUNTEER"]));
```
LenaNguyen, line 135: we also want to check `isAuthorizedByUserId` here to check that the userid in the token is the same as the userid they are signingup the shift as

LenaNguyen (changes requested):

backend/prisma/migrations/20220116032551_add_note_to_signup/migration.sql
sherryhli: With #106 just merged the latest migration timestamp is now `20220128011600` (January 28, 2022), could you please update this file name again?

backend/graphql/index.ts
hujoseph99: Added!

backend/services/implementations/shiftSignupService.ts
```diff
@@ -0,0 +1,87 @@
...
+  ): Promise<ShiftSignupResponseDTO[]> {
+    try {
+      const shiftSignups = await prisma.signup.findMany({
+        where: { userId: Number(userId) },
+      });
+      return shiftSignups.map((signup) => this.convertSignupToDTO(signup));
+    } catch (error) {
+      Logger.error(`Failed to shift signups. Reason = ${error.message}`);
+      throw error;
+    }
```
sherryhli, line 36: For type-safe error handling, please use `getErrorMessage` from utilities/errorUtils.ts

```suggestion
    } catch (error: unknown) {
      Logger.error(`Failed to shift signups. Reason = ${getErrorMessage(error)}`);
      throw error;
    }
```

backend/services/implementations/shiftSignupService.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  async createShiftSignups(
+    shiftSignups: CreateShiftSignupDTO[],
+  ): Promise<ShiftSignupResponseDTO[]> {
+    try {
+      const newShiftSignups = await prisma.$transaction(
+        shiftSignups?.map((shiftSignup) =>
```
sherryhli, line 44:
```suggestion
        shiftSignups.map((shiftSignup) =>
```

Do we need optional chaining here? `shiftSignups` should be non-null.

backend/graphql/index.ts
```diff
@@ -128,6 +132,11 @@ const authorizedByAllRoles = () =>
 const authorizedByAdmin = () => isAuthorizedByRole(new Set(["ADMIN"]));
 const authorizedByAdminAndVolunteer = () =>
   isAuthorizedByRole(new Set(["ADMIN", "VOLUNTEER"]));
+const authorizedByVolunteer = () => {
+  return (
+    isAuthorizedByRole(new Set(["VOLUNTEER"])) && isAuthorizedByUserId("userid")
```
sherryhli, line 137: I don't think `isAuthorizedByUserId` can be used as-is because it just checks the top-level argument with name `userId`, but in this case, we want a field called `userId` under an top-evel argument called `shifts` (which is an array). Would likely need a new middleware function to do this check.

@LenaNguyen Thoughts about making this a separate task as it was not included in the original ticket description?

sherryhli (changes requested):

backend/graphql/index.ts
sherryhli: We will likely need another ticket for sorting out access control anyways, we've been restricting nearly everything to admins only right now

LenaNguyen: yeah we can do it in a separate ticket

backend/prisma/schema.prisma
```diff
@@ -124,15 +124,16 @@ model Shift {
 }
 
 model Signup {
-  id            Int          @id @default(autoincrement())
   shift         Shift        @relation(fields: [shiftId], references: [id])
   shiftId       Int          @map("shifts_id")
   user          User         @relation(fields: [userId], references: [id])
```
sherryhli, line 129: Another thought: we can just make the foreign key reference on the volunteers table rather than the base users table to enforce the "only volunteers can sign up" rule

LenaNguyen: That also works but I think our middleware for checking the role is good enough for now. We would still need to go through each shift to check that the volunteerID is the same as the volunteerID associated with the user in the token.

backend/graphql/index.ts
LenaNguyen: Made a ticket: #115

backend/prisma/migrations/20220116032551_add_note_to_signup/migration.sql
hujoseph99: i believe that my latest update should already be greater than `20220128011600`

backend/services/implementations/shiftSignupService.ts
hujoseph99: Hmmm, I don't think so. I just like adding it for safety

backend/graphql/index.ts
LenaNguyen, line 137: since the `isAuthorizedByUserId` doesn't work in this case, pls remove it

LenaNguyen (changes requested):
just one small thing and then i'll approve :))

backend/services/implementations/shiftSignupService.ts
sherryhli: Hmm ok, not a big deal here but I'd rather avoid optional chaining when not necessary and have it fail in an obvious manner if an unexpected null value is passed

hujoseph99: Hmm I see your point, I'll change it

LenaNguyen (approved):
Thanks for handling the changes!

sherryhli (approved):

Issue #102, opened 2022-01-23
Admin "Select Shift Times" Calendar - Implementation


sherryhli:
## User Story and Context

As an admin (Sistering employee), I need to add the specific timeslots for shifts so volunteers can identify if they are available for the opportunity.

### Product
Please see #42. Note that the attached design in #42 is outdated, here is a more recent version, however you should confirm with @amandatdu before proceeding.

**This ticket covers the parts enclosed in the red rectangle only**:
![image.png](https://images.zenhubusercontent.com/61cddbd0ece8ea4c2aab2823/4312c3a0-aa92-46cc-bbe2-a27a5e62b76f)

**Figma link:** https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Admin-Experience?node-id=186%3A2199

Please confirm with @amandatdu on which designs to use.

### Technical
* Create both a form component (in `PostingShiftsCalendar.tsx` under `frontend/src/components/admin/posting`)

## Dev Notes
This ticket will require some investigation.

**react-grid-layout:**
https://www.npmjs.com/package/react-grid-layout

Issue #59, opened 2021-11-30
Admin create posting "Review and Post" page UI


sherryhli:
## User Story and Context

As an admin (Sistering employee), I need to publish a posting so that volunteers can view it and choose to submit their availability to it.

### Product
Please see #43.

**This ticket covers the parts enclosed in the red rectangle only**:
![Untitled.png](https://images.zenhubusercontent.com/5d93d0e226e7b90001e3f02f/7dd90259-5575-492b-879e-2cedfac76771)

**Figma link:** https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Admin-Experience?node-id=186%3A2199

Please confirm with @amandatdu on which designs to use.

### Technical
This ticket is for implementing the UI only.
* Create both a display component (in `CreatePostingReview.tsx` under `frontend/src/components/admin/posting`) and a page in which the form is mounted (located in `frontend/src/components/pages/admin/posting/CreatePostingReviewPage.tsx`)
* The page should have the route `/admin/posting/create/review`
* Data is fetched from global state using the context API. Trace through the "Edit Team" example from starter code to understand this state management pattern, here is the [starting point](https://github.com/uwblueprint/sistering/blob/main/frontend/src/App.tsx#L38-L40)

## Dev Notes

**Chakra docs for recommended components:**
* [Tag](https://chakra-ui.com/docs/data-display/tag)
* [Table](https://chakra-ui.com/docs/data-display/table)

Issue #58, opened 2021-11-30
Admin create posting "Add Shifts" form and page


sherryhli:
## User Story and Context

As an admin (Sistering employee), I need to add the specific timeslots for shifts so volunteers can identify if they are available for the opportunity.

### Product
Please see #42. Note that the attached design in #42 is outdated, here is a more recent version, however you should confirm with @amandatdu before proceeding.

**This ticket covers the parts enclosed in the red rectangle only**:
![image.png](https://images.zenhubusercontent.com/61cddbd0ece8ea4c2aab2823/52534276-257c-461c-9eb0-0dc9aec43299)

**Figma link:** https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Admin-Experience?node-id=186%3A2199

Please confirm with @amandatdu on which designs to use.

### Technical
* Create both a form component (in `CreatePostingShifts.tsx` under `frontend/src/components/admin/posting`) and mount to the corresponding page (located in `frontend/src/components/pages/admin/posting/CreatePostingShiftsPage.tsx`)
* The page should have the route `/admin/posting/create/shifts`
* Form data should be written to global state using the context API so that it can easily be shared with the "Review Posting" page. **It does not need to be sent to the backend yet.** Trace through the "Edit Team" example from starter code to understand this state management pattern, here is the [starting point](https://github.com/uwblueprint/sistering/blob/main/frontend/src/App.tsx#L38-L40)
* **Input validation occurs when user presses "Next"**

## Dev Notes

**Suggested components to use:**
![Untitled.png](https://images.zenhubusercontent.com/5d93d0e226e7b90001e3f02f/82d7c63d-8098-4abe-9570-b8bd7b4653b3)

**Chakra docs:**
* [Select](https://chakra-ui.com/docs/form/select)
* [Input](https://chakra-ui.com/docs/form/input)

For input validation, you will need to pair the components with [FormControl](https://chakra-ui.com/docs/form/form-control).

Issue #57, opened 2021-11-30
Admin create posting "Basic Info" form and page


sherryhli:
## User Story and Context

As an admin (Sistering employee), I need to add details about the posting so volunteers can browse and review the relevant details before committing availability.

### Product
Please see #41. There is a discrepancy between the original ticket requirements and the designs, to clarify: role start and end dates should be on the "Add Shifts" page instead of the "Basic Info" page as per the designs.

**This ticket covers the parts enclosed in the red rectangle only**:
![Untitled.png](https://images.zenhubusercontent.com/5d93d0e226e7b90001e3f02f/46a92c8b-6c3e-4ba9-b788-756bc3720313)

**Figma link:** https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Admin-Experience?node-id=186%3A2199

Please confirm with @amandatdu on which designs to use.

### Technical
* Create both a form component (in `CreatePostingBasicInfo.tsx` under `frontend/src/components/admin/posting`) and a page in which the form is mounted (located in `frontend/src/components/pages/admin/posting/CreatePostingBasicInfoPage.tsx`)
* The page should have the route `/admin/posting/create/basic-info`
* Form data should be written to global state using the context API so that it can easily be shared with the "Review Posting" page. **It does not need to be sent to the backend yet.** Trace through the "Edit Team" example from starter code to understand this state management pattern, here is the [starting point](https://github.com/uwblueprint/sistering/blob/main/frontend/src/App.tsx#L38-L40)
* Dynamic content (e.g. options for the branch and employee POC selectors) should be fetched from the backend. However, if the backend is not yet ready, use some hard coded values.
* **Input validation occurs when user presses "Next"**

## Dev Notes

**Suggested Chakra components to use:**
![Untitled.png](https://images.zenhubusercontent.com/5d93d0e226e7b90001e3f02f/41345294-59f8-403e-ae8f-7475395ce948)

**Chakra docs:**
* [Select](https://chakra-ui.com/docs/form/select)
* [Number Input](https://chakra-ui.com/docs/form/number-input)
* [Input](https://chakra-ui.com/docs/form/input)
* [Tag](https://chakra-ui.com/docs/data-display/tag)

For input validation, you will need to pair the components with [FormControl](https://chakra-ui.com/docs/form/form-control) or a library like **[Formik](https://chakra-ui.com/docs/form/form-control#usage-with-form-libraries)**

**Rich text support:**
[...]

PR #103, opened 2022-01-23
Create context components for posting state management

sherryhli:
## Ticket link

Partially addresses #57, #58, #102, #59

## Implementation description
* Added context for posting info
* Created example of how to write to and read from global posting state (this can be deleted after implementing the actual create posting UI)

## Steps to test
1. Run the application locally (staging backend has issues right now so cannot use deployment preview link)
```bash
docker-compose up --build
```
2. Log in and click on "Edit Posting" on the default page
3. Enter some information in the form, then verify the changes on the default page (default page is accessing data from global state). Try jumping back and forth between these pages and manipulating the data. **Note that you should not reload the app as that will clear the app state**. This means you should not refresh or navigate by manually changing the URL, instead, navigate using the "Edit Posting" and "Go Back" buttons)

<p float="left">
    Capture
    Capture1
</p>

* `frontend/src/types/PostingContextTypes.ts`: does it include all data fields collected through the posting create/edit UIs?
* Does it break any existing features?

## Checklist

### DIFF
--- a/frontend/src/components/pages/Default.tsx
+++ b/frontend/src/components/pages/Default.tsx
@@ -3,6 +3,7 @@ import { useHistory } from "react-router-dom";
 import { HStack, Tag, Text } from "@chakra-ui/react";
 
 import * as Routes from "../../constants/Routes";
+import PostingContext from "../../contexts/admin/PostingContext";
 import SampleContext from "../../contexts/SampleContext";
 
 import Logout from "../auth/Logout";
@@ -17,6 +18,50 @@ const Button = ({ text, path }: ButtonProps) => {
     <button className="btn btn-primary" onClick={navigateTo} type="button">
       {text}
     </button>
+  );
+};
+
+const PostingInfoDisplay = () => {
+  const {
+    branchId,
+    skills,
+    employees,
+    title,
+    type,
+    status,
+    description,
+    startDate,
+    endDate,
+    autoClosingDate,
+    numVolunteers,
+    recurrenceInterval,
+  } = useContext(PostingContext);
+  return (
+    <div>
+      <Text textStyle="display-large">Posting Info</Text>
+      <div>Branch ID: {branchId}</div>
+      <div>Title: {title}</div>
+      <div>Type: {type}</div>
+      <div>Status: {status}</div>
+      <div>Description: {description}</div>
+      <div>Start Date: {startDate}</div>
+      <div>End Date: {endDate}</div>
+      <div>Auto-Closing Date: {autoClosingDate}</div>
+      <div># Volunteers: {numVolunteers}</div>
+      <div>Recurrence Interval: {recurrenceInterval}</div>
+      <div>
+        Skills:{" "}
+        {skills.map(
+          (name, i) => ` ${name}${i === skills.length - 1 ? "" : ","}`,
+        )}
+      </div>
+      <div>
+        Employees:{" "}
+        {employees.map(
+          (name, i) => ` ${name}${i === skills.length - 1 ? "" : ","}`,
+        )}
+      </div>
+    </div>
   );
 };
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/pages/Default.tsx
```diff
@@ -20,6 +21,50 @@ const Button = ({ text, path }: ButtonProps) => {
...
+      <div>
+        Employees:{" "}
+        {employees.map(
+          (name, i) => ` ${name}${i === skills.length - 1 ? "" : ","}`,
+        )}
+      </div>
+    </div>
+  );
+};
+
```
sherryhli, line 67: Just code for testing and demo purposes, no need to review in depth

frontend/src/components/pages/EditPostingPage.tsx
```diff
@@ -0,0 +1,298 @@
+import React, { useContext, useState } from "react";
```
sherryhli, line 1: Likewise here for this entire file

frontend/src/constants/Routes.ts
```diff
@@ -10,6 +10,8 @@ export const DONE_RESET_PASSWORD_PAGE = "/done-reset-password";
 
 export const EDIT_TEAM_PAGE = "/edit-team";
 
+export const EDIT_POSTING_PAGE = "/edit-posting";
```
sherryhli, line 13: Redundant as we'll actually use `/admin/posting/create/shifts` for the real implementation, will entirely remove this test page once actual UI is done

sherryhli:
Strange, why haven't any of the GitHub Actions been triggered? 🤔

briantu (approved):
Didn't dive deep into the code, but I played around with editing the posting and everything LGTM! (assuming the Github Actions passed). 

Side note: I noticed there is a `Recurrence Interval` field on the posting - I don't think that is part of the posting schema?

frontend/src/reducers/PostingContextReducer.ts
```diff
@@ -0,0 +1,80 @@
...
+  PostingContextType,
+} from "../types/PostingContextTypes";
+
+// TODO: use immer to build reducer
+export default (
+  state: PostingContextType,
+  action: PostingContextAction,
+): PostingContextType => {
+  switch (action.type) {
+    case "EDIT_BRANCH_ID":
```
LenaNguyen, line 12: I think it's good practice to store the action types in a const (similar to routes) and then use those constants when dispatching the action.

Also, preface each action type with something like "ADMIN_POSTING_". This way when you're tracing the dispatched actions it'll make it less ambiguous which context it came from (even though we only have one right now).

LenaNguyen (approved):
Looks great! Just one comment to address :))

frontend/src/reducers/PostingContextReducer.ts
sherryhli: Thanks, added `ADMIN_POSTING_` prefixes! 👍

The dispatcher typechecks the action type so that ensures it must always be one of the type values in `PostingContextAction`. I think it's unnecessary to use consts in this case, thoughts?

sherryhli:
> Side note: I noticed there is a Recurrence Interval field on the posting - I don't think that is part of the posting schema?

@briantu yep that's right! A posting on the frontend does not correspond 1-to-1 with a posting on the backend, rather it's really a combination of posting + ShiftBulkRequest (which requires recurrence interval)

LenaNguyen (approved):
LGTM!

frontend/src/components/pages/Default.tsx
sherryhli: @briantu you may find it helpful to refer to the code here for your ticket on the "review and post" page 🙂

frontend/src/components/pages/EditPostingPage.tsx
```diff
@@ -0,0 +1,298 @@
...
+      value: parseInt(e.currentTarget.value, 10),
+    });
+  };
+
+  const setRecurrenceInterval = (e: React.FormEvent<HTMLInputElement>) => {
+    dispatchPostingUpdate({
+      type: "ADMIN_POSTING_EDIT_RECURRENCE_INTERVAL",
+      value: e.currentTarget.value as RecurrenceInterval,
+    });
+  };
```
sherryhli, line 147: @jiarucao this illustrates how we dispatch actions to update the posting context, feel free to refer to it while working on your shift form UI ticket 🙂

PR #111, opened 2022-01-30
Add EmployeeUser CRUD and GraphQL Resolvers

jiarucao:
## Ticket link

Addresses second half of #29 

## Implementation description
* Implement CRUD methods:
  - getEmployeeUserById
  - getEmployeeUserByEmail
  - getEmployeeUsers
  - createEmployeeUser
  - updateEmployeeUserById
  - deleteEmployeeUserById
  - deleteEmployeeUserByEmail

* Create corresponding GraphQL resolvers and DTOs

## Steps to test
1. Start docker and run `docker-compose up --build`
2. Apply the lasted migration files by running `docker exec -it sistering_backend /bin/bash` and `npx prisma migrate deploy`. 
3. We need row(s) in the posting and branch tables for testing. You can either create those using the corresponding GraphQL endpoints or use raw SQL. 
4. Go to http://localhost:5000/graphql
5. Login using an admin user 
6. Then verify that all GraphQL queries and mutations introduced in this ticket work as expected

* Verify that GraphQL endpoints work as expected

## Checklist

### DIFF
--- a/backend/graphql/types/userType.ts
+++ b/backend/graphql/types/userType.ts
@@ -29,6 +29,17 @@ const userType = gql`
     branches: [BranchResponseDTO!]!
   }
 
+  type EmployeeUserResponseDTO {
+    id: ID!
+    firstName: String!
+    lastName: String!
+    email: String!
+    role: Role!
+    phoneNumber: String
+    branchId: ID!
+    postings: [ID!]!
+  }
+
   input CreateUserDTO {
     firstName: String!
     lastName: String!
@@ -71,6 +82,25 @@ const userType = gql`
     branches: [ID!]!
   }
 
+  input CreateEmployeeUserDTO {
+    firstName: String!
+    lastName: String!
+    email: String!
+    phoneNumber: String
+    password: String!
+    branchId: ID!
+    postings: [ID!]!
+  }
+
+  input UpdateEmployeeUserDTO {
+    firstName: String!
+    lastName: String!
+    email: String!
+    phoneNumber: String
+    branchId: ID!
+    postings: [ID!]!
+  }
+
   extend type Query {
     userById(id: ID!): UserDTO!
     userByEmail(email: String!): UserDTO!
@@ -79,6 +109,9 @@ const userType = gql`
     volunteerUserById(id: ID!): VolunteerUserResponseDTO!
     volunteerUserByEmail(email: String!): VolunteerUserResponseDTO!
     volunteerUsers: [VolunteerUserResponseDTO!]!
+    employeeUserById(id: ID!): EmployeeUserResponseDTO!
+    employeeUserByEmail(email: String!): EmployeeUserResponseDTO!
+    employeeUsers: [EmployeeUserResponseDTO!]!
   }
 
   extend type Mutation {
@@ -95,6 +128,15 @@ const userType = gql`
     ): VolunteerUserResponseDTO!
     deleteVolunteerUserById(id: ID!): ID!
     deleteVolunteerUserByEmail(email: String!): ID!
+    createEmployeeUser(
+      employeeUser: CreateEmployeeUserDTO!
+    ): EmployeeUserResponseDTO!
+    updateEmployeeUserById(
+      id: ID!
+      employeeUser: UpdateEmployeeUserDTO!
+    ): EmployeeUserResponseDTO!
+    deleteEmployeeUserById(id: ID!): ID!
+    deleteEmployeeUserByEmail(email: String!): ID!
   }
 `;
 
--- a/backend/graphql/types/userType.ts
+++ b/backend/graphql/types/userType.ts
@@ -37,7 +37,6 @@ const userType = gql`
     role: Role!
     phoneNumber: String
     branchId: ID!
-    postings: [ID!]!
   }
 
   input CreateUserDTO {
@@ -89,7 +88,6 @@ const userType = gql`
     phoneNumber: String
     password: String!
     branchId: ID!
-    postings: [ID!]!
   }
 
   input UpdateEmployeeUserDTO {
@@ -98,7 +96,6 @@ const userType = gql`
     email: String!
     phoneNumber: String
     branchId: ID!
-    postings: [ID!]!
   }
 
   extend type Query {
--- a/backend/services/implementations/postingService.ts
+++ b/backend/services/implementations/postingService.ts
@@ -77,7 +77,6 @@ const convertToEmployeeResponseDTO = (
   return employees.map((employee: Employee) => {
     return {
       id: String(employee.id),
-      userId: String(employee.userId),
       branchId: String(employee.branchId),
     };
   });
--- a/backend/services/implementations/userService.ts
+++ b/backend/services/implementations/userService.ts
@@ -816,14 +816,6 @@ class UserService implements IUserService {
                 },
               },
             },
-            include: {
-              volunteer: {
-                include: {
-                  branches: true,
-                  skills: true,
-                },
-              },
-            },
           });
         } catch (postgresError: unknown) {
           const errorMessage = [
@@ -857,7 +849,6 @@ class UserService implements IUserService {
         where: { id: Number(userId) },
         include: {
           user: true,
-          postings: true,
         },
       });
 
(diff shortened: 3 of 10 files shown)


backend/graphql/types/userType.ts
```diff
@@ -71,6 +82,25 @@ const userType = gql`
...
   }
 
+  input CreateEmployeeUserDTO {
+    firstName: String!
+    lastName: String!
+    email: String!
+    phoneNumber: String
+    password: String!
+    branchId: ID!
+    postings: [ID!]!
```
sherryhli, line 92: I think employee-posting relationships are exclusively created from the posting side (meaning that when an admin creates/updates a posting, they will select some employees for it to act as POCs). As far as my understanding goes, there isn't a case when we need to manually associate an employee to some postings from the employee side. This means we do not need the `postings` field as part of the `CreateEmployeeUserDTO` and `UpdateEmployeeUserDTO` input types, nor do we need to return `postings` in the `EmployeeUserResponseDTO`.

Sorry for the extra work and not specifying this in the ticket! However, it'd be nice to not pay the performance cost of writing/reading data we don't need.

cc @JoshuaYe @LenaNguyen for confirmation

backend/services/implementations/postingService.ts
```diff
@@ -77,7 +77,6 @@ const convertToEmployeeResponseDTO = (
   return employees.map((employee: Employee) => {
     return {
       id: String(employee.id),
-      userId: String(employee.userId),
```
sherryhli, line 80: **Note to self:** possibly need a new ticket to modify `getPosting` to include all Employee fields of the point-of-contacts (e.g. email, first name, last name, etc). This is required for the posting details page in the volunteer browsing flow. Alternatively, could make it the frontend's responsiblity to fetch employee details in a separately request using IDs returned from the `getPosting` query.

Posting GraphQL queries currently return nested `EmployeeResponseDTO` objects, decide whether or not we want to change `EmployeeResponseDTO` to `EmployeeUserResponseDTO`.

cc @LenaNguyen

backend/services/implementations/userService.ts
```diff
@@ -847,6 +850,394 @@ class UserService implements IUserService {
...
     return String(user.id);
   }
+
+  async getEmployeeUserById(userId: string): Promise<EmployeeUserResponseDTO> {
+    try {
+      const employee = await prisma.employee.findUnique({
+        where: { id: Number(userId) },
+        include: {
+          user: true,
+          postings: true,
```
sherryhli, line 860: As mentioned in the GraphQL types file, we do not require `postings` info from employee queries so this `include` is not needed. Likewise for the other get functions please.

backend/services/implementations/userService.ts
```diff
@@ -847,6 +850,394 @@ class UserService implements IUserService {
...
+            role: "EMPLOYEE",
+            phoneNumber: employeeUser.phoneNumber,
+            employee: {
+              create: {
+                branch: {
+                  connect: { id: Number(employeeUser.branchId) },
+                },
+                postings: {
+                  connect: convertToNumberIds(employeeUser.postings),
+                },
```
sherryhli, line 1007: Not needed

backend/services/implementations/userService.ts
```diff
@@ -847,6 +850,394 @@ class UserService implements IUserService {
...
+                phoneNumber: employeeUser.phoneNumber,
+              },
+            },
+            branch: {
+              connect: { id: Number(employeeUser.branchId) },
+            },
+            postings: {
+              set: [],
+              connect: convertToNumberIds(employeeUser.postings),
+            },
```
sherryhli, line 1089: Not needed

backend/services/implementations/userService.ts
```diff
@@ -847,6 +850,394 @@ class UserService implements IUserService {
...
+        prisma.user.update({
+          where: {
+            id: Number(userId),
+          },
+          data: {
+            employee: {
+              update: {
+                postings: {
+                  set: [],
+                },
```
sherryhli, line 1163: This is still required though (FYI)

sherryhli (changes requested):
Looks great overall! We don't need to resolve postings when dealing with employees so just pending the removal of that code

backend/graphql/types/userType.ts
JoshuaYe: This is correct

LenaNguyen: Yup, seems right to me

backend/services/implementations/userService.ts
```diff
@@ -1,5 +1,5 @@
 import * as firebaseAdmin from "firebase-admin";
-import { PrismaClient, User, Volunteer, Skill, Branch } from "@prisma/client";
+import { PrismaClient, User, Skill, Branch, Posting } from "@prisma/client";
```
sherryhli, line 2: Nit: please remove the unused `Posting` import

backend/services/implementations/userService.ts
```diff
@@ -847,6 +842,354 @@ class UserService implements IUserService {
...
+            },
+            data: {
+              user: {
+                update: {
+                  firstName: oldEmployeeUser!.user.firstName,
+                  lastName: oldEmployeeUser!.user.lastName,
+                  role: oldEmployeeUser!.user.role,
+                  phoneNumber: oldEmployeeUser!.user.phoneNumber,
+                },
+              },
```
sherryhli, line 1088: I think the rollback should also include reverting the branch ID right?

```suggestion
              },
              branch: {
                connect: { id: Number(oldEmployeeUser!.branchId) },
              },
```

backend/services/implementations/userService.ts
```diff
@@ -847,6 +842,354 @@ class UserService implements IUserService {
...
+              phoneNumber: deletedEmployeeUser.phoneNumber,
+              employee: {
+                create: {
+                  branch: {
+                    connect: { id: Number(deletedEmployee!.branchId) },
+                  },
+                  postings: {
+                    connect: deletedEmployee!.postings.map((p) => {
+                      return { id: Number(p.id) };
+                    }),
```
sherryhli, line 1162: Hmm, I don't think this rollback is having the intended effect. `deletedEmployee` will not have any associated postings due to the `set: []` done on postings in the initial update within the transaction.

However, I think the solution is actually pretty simple 😄 Based on beahviour seen in `deletePosting` in `PostingService`, Prisma will automatically delete join table records when deleting from one of the associated relations, so the explicit `set: []` on postings isn't necessary. (In the case of postings, there are join table records for branches, skills, and employees, all of which are automatically deleted)

Let's convert that initial update to be a query instead, which fetches the employee + associated postings data prior to the delete. This gives us all the necessary fields needed to revert in the case of a firebase deletion failure. The volunteer deletion method suffers from the same issue too so let's change that as well.

backend/types.ts
```diff
@@ -75,13 +69,20 @@ export type CreateVolunteerUserDTO = Omit<VolunteerUserRequestDTO, "id"> & {
 
 export type UpdateVolunteerUserDTO = Omit<VolunteerUserRequestDTO, "id">;
 
-export type EmployeeUserDTO = UserDTO & EmployeeDTO;
+export type EmployeeDTO = {
```
sherryhli, line 72: Nit: let's call it `EmployeeUserDTO` for consistency with the other employee types

sherryhli (changes requested):
One more thing related to the deletion of both volunteers and employees, please see comment below for details

backend/services/implementations/userService.ts
jiarucao: Removed! But didn't need to add that query since the `delete` call returns the deleted object :)

backend/types.ts
```diff
@@ -75,13 +69,20 @@ export type CreateVolunteerUserDTO = Omit<VolunteerUserRequestDTO, "id"> & {
...
+
+export type EmployeeUserResponseDTO = UserDTO & EmployeeUserDTO;
 
-export type CreateEmployeeUserDTO = Omit<EmployeeUserDTO, "id"> & {
+export type CreateEmployeeUserDTO = Omit<EmployeeUserRequestDTO, "id"> & {
   password: string;
 };
 
-export type UpdateEmployeeUserDTO = Omit<EmployeeUserDTO, "id"> & {
+export type UpdateEmployeeUserDTO = Omit<EmployeeUserRequestDTO, "id"> & {
```
sherryhli, line 85: I don't think password is an updatable field (can only be reset via special reset flow), it should be omitted from the `UpdateEmployeeUserDTO` type

sherryhli (changes requested):
Two very minor things and then this is good to go!

PR #185, opened 2022-03-02
Add Support for Nested Fields in isAuthorizedByUserId for Retrieving UserId

### LINKED ISSUE #115: Middleware to check user id for nested fields
The `isAuthorizedByUserId` method in `backend\middlewares\auth.ts` should be updated to handle when the user field is nested in the request body. 

Currently, this method checks that the userID in the request body matches the userID in the request token. However, this does not work if the userID is nested.

For example, one issue we ran into is a request body like:
```
{
  shifts: [
    {
       userId: <ID>
    }
  ]
}
```
See [this comment](https://github.com/uwblueprint/sistering/pull/90#discussion_r795252688) for more context.

The function should access the shifts field and then map through each shift to check the userID field.

An alternative scenario would be:

```
{
  userData: {
       userId: 1
    }
}
```

Where the function would access the field userData and then userId.

## Dev Notes
* lodash's filter function might be useful. Example [here](https://stackoverflow.com/questions/33676823/lodash-search-nested-array-and-return-object)
    * you could fetch the user id from the token first, then use the filter function to check if the id 
       matches the expected values in the request body.

albertlai431:
## Ticket link
Closes #115

## Implementation description
* Created `isAuthorizedForCreateShiftSignups` check for `createShiftSignups` and `isAuthorizedForUpdateShiftSignups` check for `updateShiftSignup`

## Steps to test
1. Ensure that your local db contains the [seed data
](https://uwblueprint.slack.com/archives/C02CN0ZAZDG/p1645395343075529)
2. `docker-compose up --build`
3. Navigate to http://localhost:5000/graphql 
4. Run the following mutations and ensure the results are expected (Credits for the mutations to Joseph from #90 🙏).

Sign in as a volunteer and use the access token for the following mutations:
```graphql
# Create Shift Signups
mutation {
  createShiftSignups(
    shifts: [{ shiftId: 2, userId: 1, numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}
# Expected result: error with "Failed authentication and/or authorization by userId"

mutation {
  createShiftSignups(
    shifts: [{ shiftId: 2, userId: 2, numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}
# Expected result: succeed

# Update Shift Signup
mutation {
  updateShiftSignup(
    shiftId: "2"
    userId: "1"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
# Expected result: error with "Failed authentication and/or authorization by userId"

mutation {
  updateShiftSignup(
    shiftId: "2"
    userId: "2"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
# Expected result: succeed
```
Sign in as an admin and use the access token for the following mutations:
```graphql
# Create Shift Signups
mutation {
  createShiftSignups(
    shifts: [{ shiftId: 2, userId: 1, numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}
# Expected result: error with "Failed authentication and/or authorization by userId"

# Update Shift Signup
mutation {
  updateShiftSignup(
    shiftId: "2"
    userId: "2"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
# Expected result: succeed
```

Sign in as an employee and use the access token for the following mutations:
```graphql
# Create Shift Signups
mutation {
  createShiftSignups(
    shifts: [{ shiftId: 2, userId: 3, numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}
# Expected result: error with "Failed authentication and/or authorization by userId"

# Update Shift Signup
mutation {
  updateShiftSignup(
    shiftId: "2"
    userId: "3"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
# Expected result: error with "Failed authentication and/or authorization by userId"
```

* Ensuring that the mutations above produce the expected result

## Checklist

### DIFF
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -56,7 +56,7 @@ const isValidDateTime = (dateTimeString: string) => {
   return (
     !Number.isNaN(Date.parse(`${dateTimeString}:00`)) && // cover cases of DD > 31
     new Date(`${dateTimeString}:00+00:00`).toISOString().slice(0, 16) ===
-    dateTimeString
+      dateTimeString
   );
 };
 
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -3,7 +3,11 @@ import { GraphQLScalarType, Kind } from "graphql";
 import { applyMiddleware } from "graphql-middleware";
 import { merge } from "lodash";
 
-import { isAuthorizedByRole, isAuthorizedByUserId } from "../middlewares/auth";
+import {
+  isAuthorizedByRole,
+  isAuthorizedByUserId,
+  isAuthorizedByUserIdCreateShiftSignups,
+} from "../middlewares/auth";
 import authResolvers from "./resolvers/authResolvers";
 import authType from "./types/authType";
 import entityResolvers from "./resolvers/entityResolvers";
@@ -175,7 +179,7 @@ const graphQLMiddlewares = {
     updateEmployeeUserById: authorizedByAdminAndEmployee(),
     deleteEmployeeUserById: authorizedByAdmin(),
     deleteEmployeeUserByEmail: authorizedByAdmin(),
-    logout: isAuthorizedByUserId(["userId"]),
+    logout: isAuthorizedByUserId("userId"),
     createShifts: authorizedByAdmin(),
     updateShift: authorizedByAdmin(),
     updateShifts: authorizedByAdmin(),
@@ -192,9 +196,9 @@ const graphQLMiddlewares = {
     deleteBranch: authorizedByAdmin(),
     createShiftSignups:
       authorizedByVolunteer() &&
-      isAuthorizedByUserId(["shifts", "0", "userId"]),
+      isAuthorizedByUserIdCreateShiftSignups("userId"),
     updateShiftSignup:
-      authorizedByAdminAndVolunteer() && isAuthorizedByUserId(["userId"]),
+      authorizedByAdminAndVolunteer() && isAuthorizedByUserId("userId"),
   },
 };
 
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -51,7 +51,7 @@ const isValidDateTime = (dateTimeString: string) => {
   return (
     !Number.isNaN(Date.parse(`${dateTimeString}:00`)) && // cover cases of DD > 31
     new Date(`${dateTimeString}:00+00:00`).toISOString().slice(0, 16) ===
-      dateTimeString
+    dateTimeString
   );
 };
 
@@ -175,7 +175,7 @@ const graphQLMiddlewares = {
     updateEmployeeUserById: authorizedByAdminAndEmployee(),
     deleteEmployeeUserById: authorizedByAdmin(),
     deleteEmployeeUserByEmail: authorizedByAdmin(),
-    logout: isAuthorizedByUserId("userId"),
+    logout: isAuthorizedByUserId(["userId"]),
     createShifts: authorizedByAdmin(),
     updateShift: authorizedByAdmin(),
     updateShifts: authorizedByAdmin(),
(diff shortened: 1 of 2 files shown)


backend/graphql/index.ts
```diff
@@ -190,8 +190,11 @@ const graphQLMiddlewares = {
     createBranch: authorizedByAdmin(),
     updateBranch: authorizedByAdmin(),
     deleteBranch: authorizedByAdmin(),
-    createShiftSignups: authorizedByVolunteer(),
-    updateShiftSignup: authorizedByAdmin(),
+    createShiftSignups:
+      authorizedByVolunteer() &&
+      isAuthorizedByUserId(["shifts", "0", "userId"]),
```
LenaNguyen, line 195: Ideally we wanna check that every shift, not just the first one, is for the user making the request so we would have to loop through the shifts. This is why you may have to create a new function for this.

LenaNguyen (changes requested):

backend/middlewares/auth.ts
```diff
@@ -81,6 +81,43 @@ export const isAuthorizedByUserId = (userIdField: string) => {
...
+      context: ExpressContext,
+      info: GraphQLResolveInfo,
+    ) => any,
+    parent: any,
+    args: { [key: string]: any },
+    context: ExpressContext,
+    info: GraphQLResolveInfo,
+  ): Promise<any> => {
+    const accessToken = getAccessToken(context.req);
+    args.shifts.forEach(async (shift: any) => {
```
LenaNguyen, line 102: you can't use `await` inside a forEach function in javascript. (see [here](https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop)) Try using map and awaiting the result instead. Otherwise, this function exits before the shift ids are actually checked.

LenaNguyen (changes requested):
Not sure if this is just me, but it seems like the authorization checks aren't working now? I was able to create a shift for any user but also with any role 🤔 Not sure what's happening yet

backend/middlewares/auth.ts
```diff
@@ -81,6 +81,93 @@ export const isAuthorizedByUserId = (userIdField: string) => {
...
+      info: GraphQLResolveInfo,
+    ) => any,
+    parent: any,
+    args: { [key: string]: any },
+    context: ExpressContext,
+    info: GraphQLResolveInfo,
+  ): Promise<any> => {
+    const accessToken = getAccessToken(context.req);
+    const authorizedArray = await Promise.all(
+      args.shifts.map(async (shift: any) => {
```
LenaNguyen, line 103: You could use `shifts.every()` instead of map to check that every value returned is true and then store the result in a variable like `isAuthorized`. This way you won't need to iterate through the array again in `authorizedArray.includes()`.

backend/middlewares/auth.ts
```diff
@@ -81,6 +81,93 @@ export const isAuthorizedByUserId = (userIdField: string) => {
...
+        if (accessToken) {
+          return (
+            (await authService.isAuthorizedByUserId(
+              accessToken,
+              shift[userIdField],
+            )) &&
+            (await authService.isAuthorizedByRole(
+              accessToken,
+              new Set(["VOLUNTEER"]),
+            ))
```
LenaNguyen, line 113: I believe we only need to check the role once. Let's move this to before we iterate through all the shifts. If `isAuthorizedByRole` returns false, then throw the `AuthenticationError` immediately after so we don't need to spend time iterating through shifts if not needed.

backend/middlewares/auth.ts
```diff
@@ -81,6 +81,93 @@ export const isAuthorizedByUserId = (userIdField: string) => {
...
+    const accessToken = getAccessToken(context.req);
+    const authorized =
+      accessToken &&
+      ((await authService.isAuthorizedByRole(
+        accessToken,
+        new Set(["ADMIN"]),
+      )) ||
+        ((await authService.isAuthorizedByRole(
+          accessToken,
+          new Set(["VOLUNTEER"]),
```
LenaNguyen, line 155: This chain is hard to read. Can you store the result of this role check in a variable called `roleIsAuthorized`  ? 

So then `const isAuthorized = roleIsAuthorized&& (await authService.isAuthorizedByUserId(....)`

LenaNguyen (changes requested):
Everything works great now! Thank you for looking into this! Just some tips on optimizing the code but aside from that it's looking good :)

backend/middlewares/auth.ts
LenaNguyen: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every

LenaNguyen (approved):
Thanks for incorporating the changes. This looks great!

PR #232, opened 2022-03-24
Skip shifts that occur before start date

LenaNguyen:
## Ticket link

Closes #

## Implementation description
* 

## Steps to test
Example using graphql:
```
query Postings {
  postings {
    id
    title
    description
    startDate
    endDate
    shifts {
      startTime
      endTime
    }
  }
}

mutation CreateShifts{
  createShifts(shifts:{postingId:<POSTING_ID>, times:[{startTime:"2022-03-28T01:30", endTime:"2022-03-28T02:00"},{startTime:"2022-03-29T01:30", endTime:"2022-03-29T02:00"} ], startDate:"2022-03-29", endDate:"2022-03-31" recurrenceInterval:NONE}) {
    id
    startTime
    endTime
  }
}
```

Through the UI:
Select "Create Posting" in the navbar on the homepage.
When you get to the page for creating shifts, choose a start date that does not begin on sunday. Add a shift that begins before the start date. 

When submitting the posting, there should be no errors. Check through graphql (using the Postings query above) that shifts before the start date were not generated.

*  Shifts before the start date are skipped while shifts after the start date are not

## Checklist


backend/services/implementations/shiftService.ts
```diff
@@ -111,27 +111,30 @@ class ShiftService implements IShiftService {
...
-    const shiftTimes: TimeBlock[] = shifts.times.flatMap((time) => {
-      const recurringShifts = [];
-      if (interval > 0) {
-        for (
-          let start = time.startTime.getTime(), end = time.endTime.getTime();
-          start < endDate.getTime();
-          start += interval, end += interval
-        ) {
+    const shiftTimes: TimeBlock[] = shifts.times.flatMap(
+      ({ startTime, endTime }) => {
```
LenaNguyen, line 115: just destructured `time` here. No functional changes

LenaNguyen:
Should we create a new endpoint that creates a posting and its shifts? If shifts throw an error then I would imagine that we don't want to create the posting anymore

backend/services/implementations/shiftService.ts
```diff
@@ -225,12 +228,21 @@ class ShiftService implements IShiftService {
 
   async createShifts(shifts: ShiftBulkRequestDTO): Promise<ShiftResponseDTO[]> {
     try {
+      const filteredShifts = shifts;
+
+      // Skip shifts that occur before start date
+      filteredShifts.times = shifts.times.filter(
+        (shift) => shift.startTime.getTime() > shifts.startDate.getTime(),
```
sherryhli, line 235: Should this be `>=`?

sherryhli (commented):
Only reviewed the last commit (7f2a87f3e20db5d80f0108aaf96e70c869819fa7)

sherryhli (approved):
🚢🚢🚢

> Should we create a new endpoint that creates a posting and its shifts? If shifts throw an error then I would imagine that we don't want to create the posting anymore

Missed this earlier but yeah, I agree it should be all or nothing. Made a ticket for (#244) that we should get assigned out for next sprint so that it can be included for soft launch

PR #253, opened 2022-04-01
Add submit signups mutation to availability page

### LINKED ISSUE #230: Volunteer Availabilities submit signups mutation call

### LINKED ISSUE #258: Create a `upsertDeleteSignupShifts` mutation to overwrite all signups for user
### Context

When a user submits their availabilities, there are 3 types of changes that they could be making:
* Creating new signup for a shift (adding a brand new shift-user mapping in `signups` table)
* Updating an existing signup (for example, editing the note)
* Deleting a previous signup

Currently we only support the first case.

### Acceptance Criteria
Create a new mutation that will upsert and delete shift signups for a given userId and postingId. This mutation will essentially replace the user's signups for the posting with the array of signups provided in the request.
* For the list of `signup` objects supplied in the request, perform an **upsert** in the database rather than a pure insert
* In your deleteMany query, any signupIds not in the list signups should be deleted. 

### Technical Notes
* This is the suggested Prisma operation to achieve the desired behaviour: https://github.com/prisma/prisma/issues/2255#issuecomment-683811551
* If the above doesn't work, we will need to write our own version (within a transaction, first delete the `toDelete`s, then upsert)

[^1]: Should be renamed

MatoPlus:
## Ticket link

Closes #258

## Implementation description
* Add mutation call that does delete then upsert on signup shifts
* New mutation call has new types added

NOTE: Only the backend mutation is completed, not integrated into client at the moment.

## Steps to test
1. Create new signups
2. Use the new mutation call to update and delete some signups in the same graphql mutation

* Correctness
* New Typing

## Checklist

### DIFF
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -198,6 +198,7 @@ const graphQLMiddlewares = {
     deleteBranch: authorizedByAdmin(),
     createShiftSignups: isAuthorizedForCreateShiftSignups("userId"),
     updateShiftSignup: isAuthorizedForUpdateShiftSignups("userId"),
+    upsertDeleteShiftSignups: isAuthorizedForUpdateShiftSignups("userId"),
   },
 };
 
--- a/backend/graphql/index.ts
+++ b/backend/graphql/index.ts
@@ -198,7 +198,7 @@ const graphQLMiddlewares = {
     deleteBranch: authorizedByAdmin(),
     createShiftSignups: isAuthorizedForCreateShiftSignups("userId"),
     updateShiftSignup: isAuthorizedForUpdateShiftSignups("userId"),
-    upsertDeleteShiftSignups: isAuthorizedForUpdateShiftSignups("userId"),
+    upsertDeleteShiftSignups: authorizedByAdminAndVolunteer(),
   },
 };
 
--- a/backend/services/implementations/shiftSignupService.ts
+++ b/backend/services/implementations/shiftSignupService.ts
@@ -6,6 +6,7 @@ import {
   CreateShiftSignupDTO,
   ShiftSignupResponseDTO,
   UpdateShiftSignupRequestDTO,
+  UpsertDeleteShiftSignupsRequestDTO,
 } from "../../types";
 import logger from "../../utilities/logger";
 import { getErrorMessage } from "../../utilities/errorUtils";
@@ -150,6 +151,65 @@ class ShiftSignupService implements IShiftSignupService {
     } catch (error: unknown) {
       Logger.error(
         `Failed to update shift signup. Reason = ${getErrorMessage(error)}`,
+      );
+      throw error;
+    }
+  }
+
+  async upsertDeleteShiftSignups(
+    upsertDeleteShiftSignups: UpsertDeleteShiftSignupsRequestDTO,
+  ): Promise<ShiftSignupResponseDTO[]> {
+    try {
+      const newShiftSignups = await prisma.$transaction([
+        prisma.signup.deleteMany({
+          where: {
+            OR: upsertDeleteShiftSignups.deleteShiftSignups.map(
+              (deleteShiftSignup) => ({
+                shiftId: Number(deleteShiftSignup.shiftId),
+                userId: Number(deleteShiftSignup.userId),
+              }),
+            ),
+          },
+        }),
+        ...upsertDeleteShiftSignups.upsertShiftSignups.map(
+          (upsertShiftSignup) => {
+            const { shiftId, userId, status, ...signup } = upsertShiftSignup;
+            return prisma.signup.upsert({
+              where: {
+                shiftId_userId: {
+                  shiftId: Number(upsertShiftSignup.shiftId),
+                  userId: Number(upsertShiftSignup.userId),
+                },
+              },
+              create: {
+                ...signup,
+                status: SignupStatus.PENDING,
+                shiftId: Number(shiftId),
+                userId: Number(userId),
+              },
+              update: {
+                ...signup,
+                shiftId: Number(shiftId),
+                userId: Number(userId),
+                ...(status === null ? {} : { status }),
+              },
+              include: {
+                shift: true,
+              },
+            });
+          },
+        ),
+      ]);
+
+      return newShiftSignups.map((newShiftSignup) => {
+        const signup = newShiftSignup as Signup & {
+          shift: Shift;
+        };
+        return this.convertSignupResponseToDTO(signup, signup.shift);
+      });
+    } catch (error: unknown) {
+      Logger.error(
+        `Failed to create shift signup. Reason = ${getErrorMessage(error)}`,
       );
       throw error;
     }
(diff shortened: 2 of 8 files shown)


backend/graphql/index.ts
```diff
@@ -198,6 +198,9 @@ const graphQLMiddlewares = {
     deleteBranch: authorizedByAdmin(),
     createShiftSignups: isAuthorizedForCreateShiftSignups("userId"),
     updateShiftSignup: isAuthorizedForUpdateShiftSignups("userId"),
+    upsertDeleteShiftSignups:
+      isAuthorizedForCreateShiftSignups("userId") &&
```
MatoPlus, line 202: @uwblueprint/sistering-w22-pls can I get a confirmation that this check is right? I assume to delete, you need perm to create + update (counts as delete as well)

backend/services/implementations/shiftSignupService.ts
```diff
@@ -154,6 +155,65 @@ class ShiftSignupService implements IShiftSignupService {
...
+                shiftId: Number(shiftId),
+                userId: Number(userId),
+                ...(status === null ? {} : { status }),
+              },
+              include: {
+                shift: true,
+              },
+            });
+          },
+        ),
```
MatoPlus, line 201: In order to make this code robust, I don't think I can do this query in one `prisma.x.update` call. main reason being that to do so, we need to ensure that userId is the same for all shiftsign ups, I don't think there's really any guarantee that they have the same userId, so this can be used a general backend query for combining delete +upserts

I think this is a bit more readable as well since it's just subsequent calls in a tcx

frontend/src/components/pages/volunteer/posting/VolunteerPostingAvailabilities.tsx
```diff
@@ -85,10 +102,23 @@ const VolunteerPostingAvailabilities = (): React.ReactElement => {
...
             console.log(shiftSignups);
             console.log(signupNotes);
+
+            // Assume Num volunteers would be one...
+            // TODO: Merge data from shiftSignups and signupNotes together to pass into query
+            const graphQLResult = await submitSignups({
+              variables: {
+                upsertDeleteShifts: {
+                  upsertShiftSignups: [],
+                  deleteShiftSignups: [],
```
MatoPlus, line 116: This part isn't trivial, i feel like our client types isn't super clean atm.

backend/graphql/index.ts
LenaNguyen: isAuthorizedForUpdateShiftSignups should be enough because it already includes the same permissions as the createShiftSignups.

Basically you can only create a signup if you're a volunteer and your userId matches the userId of the signups.

You can update a singup if you match the above OR if you're an admin (so we can update the shift status)

backend/services/implementations/shiftSignupService.ts
```diff
@@ -154,6 +155,65 @@ class ShiftSignupService implements IShiftSignupService {
...
+      ]);
+
+      return newShiftSignups.map((newShiftSignup) => {
+        const signup = newShiftSignup as Signup & {
+          shift: Shift;
+        };
+        return this.convertSignupResponseToDTO(signup, signup.shift);
+      });
```
MatoPlus, line 209: Since we're using a mixed transaction of different ops, I don't think we get a clean response from prisma, this is wrong. with empty response, I got something like `[ { count: 0 } ]`. I'm investigating options.

backend/services/implementations/shiftSignupService.ts
```diff
@@ -154,6 +155,65 @@ class ShiftSignupService implements IShiftSignupService {
...
+    upsertDeleteShiftSignups: UpsertDeleteShiftSignupsRequestDTO,
+  ): Promise<ShiftSignupResponseDTO[]> {
+    try {
+      const newShiftSignups = await prisma.$transaction([
+        prisma.signup.deleteMany({
+          where: {
+            OR: upsertDeleteShiftSignups.deleteShiftSignups.map(
+              (deleteShiftSignup) => ({
+                shiftId: Number(deleteShiftSignup.shiftId),
+                userId: Number(deleteShiftSignup.userId),
```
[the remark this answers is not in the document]
MatoPlus: 100%. Good catch. I'll change the predicate to search for all conjugations of shiftid/userid that makes up the pid for signups.

LenaNguyen (commented):
nvm 😅

backend/graphql/index.ts
```diff
@@ -198,6 +198,7 @@ const graphQLMiddlewares = {
     deleteBranch: authorizedByAdmin(),
     createShiftSignups: isAuthorizedForCreateShiftSignups("userId"),
     updateShiftSignup: isAuthorizedForUpdateShiftSignups("userId"),
+    upsertDeleteShiftSignups: isAuthorizedForUpdateShiftSignups("userId"),
```
LenaNguyen, line 201: Looking at this again, we will actually need a custom authorization method similar to `isAuthorizedForCreateShiftSignups`. Currently this endpoint should only be accessible to volunteers. We would need to loop through each shift in  upsertShiftSignups and DeleteShiftSignups and check that the userID matches the userID of the accessToken.

This can be done in a different ticket. For now let's just use `authorizedByAdminAndVolunteer` to make testing easy.

frontend/src/components/pages/volunteer/posting/VolunteerPostingAvailabilities.tsx
```diff
@@ -1,13 +1,16 @@
 import React, { useState } from "react";
 import { Box, Button, Flex, Spacer, Spinner, Text } from "@chakra-ui/react";
 
-import { gql, useQuery } from "@apollo/client";
+import { gql, useMutation, useQuery } from "@apollo/client";
```
LenaNguyen, line 4: can you remove the frontend changes from this pr? let's just keep it to backend to match the scope of the ticket.

backend/services/implementations/shiftSignupService.ts
```diff
@@ -154,6 +155,68 @@ class ShiftSignupService implements IShiftSignupService {
       throw error;
     }
   }
+
+  async upsertDeleteShiftSignups(
+    upsertDeleteShiftSignups: UpsertDeleteShiftSignupsRequestDTO,
```
LenaNguyen, line 160: can the param just be deleteShiftSignups and upsertShiftSignups instead of putting them in an object?

LenaNguyen (dismissed):
It's working great! Just a couple comments

frontend/src/components/pages/volunteer/posting/VolunteerPostingAvailabilities.tsx
MatoPlus: for sure, I'll keep the revert commit in the merge for reference

LenaNguyen (changes requested):
LGTM! Just run `yarn fix` for the linting warning

LenaNguyen (commented):
TO DO: remove createShiftSignup and updateShiftSignup

LenaNguyen (approved):

PR #259, opened 2022-04-02
Schedule Review Table Data Fetch

### LINKED ISSUE #162: Schedule Review Table Data Fetch
Fetch the data needed for the Schedule Review Table and display it in the UI made in #158.

You will need to query for `shiftSignups` for a given posting, where the status is `SignupStatus.CONFIRMED`

hujoseph99:
## Ticket link

Closes #162

## Implementation description
* Created new endpoint `shiftsWithSignupsAndVolunteersByPosting`
* This is a draft PR, still have to add the front end data-fetching 

## Steps to test
1. Populate database as in this #90 
2. Add skills and branches for the user with the following queries: `
  * `insert into skills values (1, "test skill");`
  * `insert into branches values (1, "test branch");`
  * `insert into "_SkillToVolunteer" values (1, 1);`
  * `insert into  "_BranchToVolunteer" values (1, 1);`
3. Test the query to see if it works as expected in `:5000/graphql`
4. Go to frontend route `/admin/scheduleTableDemo` to see demo of query
5. Input appropriate posting ID, user ID, and signup status for testing

Note that the queries `createShifts`, `createShiftSignups`, and `updateShiftSignup` will likely be useful for creating shifts and signing up to them.

* The reviewer should focus on checking to see if the table is populated as expected with the different filters for posting ID, user ID, and signup status.

## Checklist


backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+    return Promise.all(
+      shifts.map(async (shift) => {
+        const filteredSignups: SignupsAndVolunteerResponseDTO[] = [];
+        await Promise.all(
+          shift.signups.map(async (signup: SignupWithVolunteers) => {
+            try {
+              if (
+                (userId == null || signup.userId === Number(userId)) &&
+                (signupStatus == null || signup.status === signupStatus)
+              ) {
```
LenaNguyen, line 322: you don't need this if statement if you add filtering on signups to the prisma call right?

backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+                  .getUser(signup.user.authId);
+
+                filteredSignups.push(
+                  this.convertSignupResponeWithUserAndVolunteerToDTO(
+                    signup,
+                    shift.startTime,
+                    shift.endTime,
+                    firebaseUser.email ?? "",
+                  ),
+                );
```
LenaNguyen, line 334: No action needed on this pr but we should probably store user emails in the db so we don't have to make a call to firebase every time. We'll do this in the future!

backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+        await Promise.all(
+          shift.signups.map(async (signup: SignupWithVolunteers) => {
+            try {
+              if (
+                (userId == null || signup.userId === Number(userId)) &&
+                (signupStatus == null || signup.status === signupStatus)
+              ) {
+                const firebaseUser = await firebaseAdmin
+                  .auth()
+                  .getUser(signup.user.authId);
```
LenaNguyen, line 325: To make this a bit faster, if `userId != null` then we only need to fetch the user email once since all signups will be for the same user.

backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+    } catch (error: unknown) {
+      Logger.error(`Failed to get shifts. Reason = ${getErrorMessage(error)}`);
+      throw error;
+    }
+
+    return Promise.all(
+      shifts.map(async (shift) => {
+        const filteredSignups: SignupsAndVolunteerResponseDTO[] = [];
+        await Promise.all(
+          shift.signups.map(async (signup: SignupWithVolunteers) => {
```
LenaNguyen, line 317: Do you need two `Promise.all` ? Not sure if this works but can you do `await shift.signups.map(...)` inside the first promise.all?

LenaNguyen (changes requested):
looking good, the main changes have to do with 1) removing skills + branches from the response and 2) add another `where` filter to the prisma query for `userId` and `signupStatus`

backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+  ): Promise<ShiftWithSignupAndVolunteerResponseDTO[]> {
+    let shifts: ShiftWithSignupAndVolunteers[] = [];
+    try {
+      shifts = await prisma.shift.findMany({
+        where: {
+          postingId: Number(postingId),
+        },
+        include: {
+          signups: {
+            include: {
```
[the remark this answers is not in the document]
hujoseph99: It actually isn't possible to do this in prisma as far as I understand because `signups` is a list. The closest equivalent prisma has is `has`, but that only works on scalar lists.

backend/services/implementations/shiftService.ts
hujoseph99: Correct, but I don't think this type of filtering exists in Prisma? I could be wrong though

backend/services/implementations/shiftService.ts
hujoseph99: I tried to remove the inner one initially, but it doesn't work because the `await` is container within the inner map. I just realized that I might be able to remove the outer `Promise.all` though. Let me try it out!

backend/services/implementations/shiftService.ts
```diff
@@ -219,6 +277,81 @@ class ShiftService implements IShiftService {
...
+              user: {
+                include: {
+                  volunteer: {
+                    include: {
+                      branches: true,
+                      skills: true,
```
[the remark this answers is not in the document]
hujoseph99: Removed

backend/services/implementations/shiftService.ts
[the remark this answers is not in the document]
LenaNguyen: hmm could you use [every](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#relation-filters)?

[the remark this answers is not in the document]
hujoseph99: I've tested it and it unfortunately does not work. I believe it only matches shift signups where **every** volunteer in the array matched our filters. So, if we're filtering by userId and a posting has signups with two different userIds, it just won't return that posting since not every element in the array has matched our filters.

backend/services/implementations/shiftService.ts
```diff
@@ -159,6 +191,32 @@ class ShiftService implements IShiftService {
...
+      shiftId: String(signup.shiftId),
+      shiftStartTime,
+      shiftEndTime,
+      volunteer: {
+        ...signup.user,
+        ...signup.user.volunteer,
+        id: String(signup.user.id),
+        email,
+        branches: convertToBranchResponseDTO(signup.user.volunteer.branches),
+        skills: convertToSkillResponseDTO(signup.user.volunteer.skills),
```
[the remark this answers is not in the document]
hujoseph99: Removed

backend/services/implementations/shiftService.ts
hujoseph99: As discussed, we removed emails from the response in this endpoint

backend/services/implementations/shiftService.ts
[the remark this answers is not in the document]
LenaNguyen: I tried 

```suggestion
        include: {
          signups: {
             where: {
              userId: userId === undefined ? userId : Number(userId),
              status: signupStatus ?? undefined,
            },
            include: {
```
and removed the filter we do after. It seems to work for me. Can you try it out?

LenaNguyen (changes requested):

backend/services/implementations/shiftService.ts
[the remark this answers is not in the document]
LenaNguyen: Testing for getShiftsWithSignupAndVolunteerForPosting :

1. Go to backend\graphql\index.ts. On line 200 - 201, change the auth for createShiftSignups and updateShiftSignup to be authorizedByAdminAndVolunteer() to make testing easier.

2. Create a new user

3. Go into the db and update the user role to be volunteer + insert a new volunteer:
`update users set role = 'VOLUNTEER' where id = 4;`
`insert into volunteers(id, hire_date, date_of_birth, pronouns) values(4, '2022-04-17 19:41:55.55', '2000-08-19 23:15:30', 'they/them');`

Create shift signups for both volunteers:
```
mutation CreateShiftSignups{
  createShiftSignups(
    shifts: [ { shiftId: "5", userId: "4", numVolunteers: 3, note: "test" }, { shiftId: "2", userId: "4", numVolunteers: 3, note: "test" }, { shiftId: "2", userId: "2", numVolunteers: 3, note: "test" }, { shiftId: "1", userId: "2", numVolunteers: 3, note: "test" }]
  ) {
    shiftId
    userId
    numVolunteers
    note
  }
}
```

4. Update at least one shift signup to have a status other than pending:
```
mutation UpdateShiftSignup{
  updateShiftSignup(
    shiftId: "1"
    userId: "2"
  ) {
    shiftId
    userId
    numVolunteers
    note
    status
  }
}
```

5. Query for shifts with signups and volunteers using various filter combinations (e.g. just posting id, posting id  + user id, posting id + user id + signup status)
```
query ShiftsWithSignupVolunteers {
  shiftsWithSignupsAndVolunteersByPosting(postingId:1, userId: 4, signupStatus:PENDING) {
    id
    postingId
    startTime
    endTime
    signups {
      status
      shiftId
      userId
    }
  }
}
```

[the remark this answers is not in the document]
hujoseph99: This seems to work! I've made the changes and gone with this approach now. Thanks for the testing :)

frontend/src/types/api/UserType.ts
```diff
@@ -0,0 +1,24 @@
+import { BranchResponseDTO } from "./BranchTypes";
+import { SkillResponseDTO } from "./SkillTypes";
+
+export type Role = "ADMIN" | "VOLUNTEER" | "EMPLOYEE";
+
+export type UserDTO = {
+  id: string;
+  firstName: string;
+  lastName: string;
+  email: string;
```
LenaNguyen, line 10: remove email since we're not returning it anymore

hujoseph99: removed

LenaNguyen (approved):
Thanks for incorporating all the changes! So glad to have this merged in :))

PR #261, opened 2022-04-07
Load initial create posting values from context

### LINKED ISSUE #233: Load initial values for create posting pages from context
Follows #138

### Acceptance Criteria
Load initial values from context so that data is not lost on navigating to "back" and "next"[^1]

[^1]: This likely requires refactoring how we store the selected branch, skills, and employees in the component state. They currently depend on loading options from backend and will cause errors while waiting for the API call to return. Potentially need to delay rendering until API call returns? Loading state design is needed in this case.

sherryhli:
## Ticket link

Closes #233 

## Implementation description
* Set initial values in create basic info and shift pages to be values from the posting context
* Note context is updated on "Next", so previous values can only be retrieved if "Next" was clicked prior to navigating away

## Steps to test
1. Click preview deploy link
2. Navigate to create posting page
3. Enter info on each page and test navigating forward/back with the "Next" and "Back" buttons, the previously entered data should be preserved

* Does it work?
* See comments

## Checklist

### DIFF
--- a/frontend/src/components/admin/posting/CreatePostingBasicInfo.tsx
+++ b/frontend/src/components/admin/posting/CreatePostingBasicInfo.tsx
@@ -33,6 +33,7 @@ import {
   ADMIN_POSTING_CREATE_BASIC_INFO_ROLE_DESCRIPTION_PLACEHOLDER,
   ADMIN_POSTING_CREATE_BASIC_INFO_SKILLS_TOOLTIP,
 } from "../../../constants/Copy";
+import PostingContext from "../../../contexts/admin/PostingContext";
 import PostingContextDispatcherContext from "../../../contexts/admin/PostingContextDispatcherContext";
 
 import { BranchDTO, BranchResponseDTO } from "../../../types/api/BranchTypes";
@@ -73,6 +74,15 @@ const ERROR_MESSAGE_HEIGHT = "35px";
 const CreatePostingBasicInfo: React.FC<CreatePostingBasicInfoProps> = ({
   navigateToNext,
 }: CreatePostingBasicInfoProps): React.ReactElement => {
+  const {
+    branch: branchFromCtx,
+    skills: skillsFromCtx,
+    employees: employeesFromCtx,
+    title: titleFromCtx,
+    description: descriptionFromCtx,
+    autoClosingDate: autoClosingDateFromCtx,
+    numVolunteers: numVolunteersFromCtx,
+  } = useContext(PostingContext);
   const dispatchPostingUpdate = useContext(PostingContextDispatcherContext);
 
   // #region state variables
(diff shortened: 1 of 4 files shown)


frontend/src/components/admin/posting/CreatePostingBasicInfo.tsx
```diff
@@ -445,7 +462,8 @@ const CreatePostingBasicInfo: React.FC<CreatePostingBasicInfoProps> = ({
                       mr={4}
                       mb={2}
                     >
-                      {getOptionNameFromId(skillOptions, skillId)}
+                      {skillOptions.length > 0 &&
+                        getOptionNameFromId(skillOptions, skillId)}
```
sherryhli, line 466: Looking for a second opinion on this:

* Basic info form stores selected branch/skills/employees by their IDs only in state and looks up the rest of the fields using the GraphQL query responses for branch/skills/employees. The result of this lookup is added to the context
* Context stores full branch/skills/employee objects because fields such as skill name, employee email, etc. need to be displayed on review page

**Advantage:** ensures mostly update to date values for the additional fields, review posting page does not need to make additional API calls to get the extra fields
**Drawback:** If branch/skill/employee options are removed in the backend between the time selections are first added to context and page is visited again, there may not be valid mappings between old choices and current options, this is an unhandled error case. However, I think this is an exceptionally rare case because these options hardly ever change

frontend/src/contexts/admin/PostingContext.ts
```diff
@@ -24,7 +24,7 @@ export const DEFAULT_POSTING_CONTEXT = {
       endTime: "2022-02-06T13:30",
     },
   ],
-  recurrenceInterval: "NONE" as RecurrenceInterval,
+  recurrenceInterval: "" as RecurrenceInterval,
```
sherryhli, line 27: Thoughts on handling the default value? I don't like using type assertion here as `""` is not a valid recurrence interval, but also don't want to make `recurrenceInterval` nullable just for this case

LenaNguyen (approved):
Just left a suggestion about the type! I don't think it's too big of an issue so LGTM but if you decide to change it, I'll reapprove :)

frontend/src/components/admin/posting/CreatePostingBasicInfo.tsx
sherryhli: Yeah, agree with that. I'll make note in the current error state ticket

frontend/src/contexts/admin/PostingContext.ts
sherryhli: Done, thanks for the suggestion!

LenaNguyen (approved):

PR #273, opened 2022-04-17
Create shifts with posting

### LINKED ISSUE #244: Create posting and associated shifts as single unit of work in backend
Should we create a new endpoint that creates a posting and its shifts? If shifts throw an error then I would imagine that we don't want to create the posting anymore

_Originally posted by @LenaNguyen in https://github.com/uwblueprint/sistering/issues/232#issuecomment-1077778349_

LenaNguyen:
## Ticket link

Closes #244 

## Implementation description
* Created a new method called `bulkGenerateTimeBlocks` that will verify and create an array of TimeBlocks to be used creating a posting
* `CreatePosting` now generates shifts along with the posting
* Front end of the Create posting review page has been updated to match the modified endpoint

## Steps to test
1. Create a new posting (try recurring and non-recurring shifts) through the UI
2. After submitting, query for the posting and check that it was created properly with the associated shifts

```
query Postings {
  postings {
    id
    title
    description
    startDate
    endDate
    status
    shifts {
      startTime
      endTime
    }
  }
}
```

*  Shifts are generated with the same logic as before

## Checklist

### DIFF
--- a/backend/graphql/resolvers/postingResolvers.ts
+++ b/backend/graphql/resolvers/postingResolvers.ts
@@ -1,8 +1,15 @@
 import PostingService from "../../services/implementations/postingService";
 import IPostingService from "../../services/interfaces/postingService";
-import { PostingRequestDTO, PostingResponseDTO } from "../../types";
+import {
+  PostingRequestDTO,
+  PostingResponseDTO,
+  PostingWithShiftsRequestDTO,
+} from "../../types";
+import IShiftService from "../../services/interfaces/IShiftService";
+import ShiftService from "../../services/implementations/shiftService";
 
-const postingService: IPostingService = new PostingService();
+const shiftService: IShiftService = new ShiftService();
+const postingService: IPostingService = new PostingService(shiftService);
 
 const postingResolvers = {
   Query: {
@@ -19,7 +26,7 @@ const postingResolvers = {
   Mutation: {
     createPosting: async (
       _parent: undefined,
-      { posting }: { posting: PostingRequestDTO },
+      { posting }: { posting: PostingWithShiftsRequestDTO },
     ): Promise<PostingResponseDTO> => {
       const newPosting = await postingService.createPosting(posting);
       return newPosting;
--- a/backend/services/implementations/postingService.ts
+++ b/backend/services/implementations/postingService.ts
@@ -14,11 +14,13 @@ import {
   PostingResponseDTO,
   PostingStatus,
   PostingType,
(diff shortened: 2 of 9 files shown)


backend/types.ts
```diff
@@ -116,8 +116,15 @@ export type ShiftBulkRequestDTO = {
   recurrenceInterval: RecurrenceInterval;
 };
 
+export type ShiftDataWithoutPostingId = Omit<ShiftBulkRequestDTO, "postingId">;
```
sherryhli, line 119: Feel free to reword but I suggest something along the lines of this so that there no confusion on whether the type refers to a single shift or not :)

```suggestion
export type ShiftBulkRequestWithoutPostingId = Omit<ShiftBulkRequestDTO, "postingId">;
```

backend/graphql/resolvers/postingResolvers.ts
```diff
@@ -19,7 +26,7 @@ const postingResolvers = {
   Mutation: {
     createPosting: async (
       _parent: undefined,
-      { posting }: { posting: PostingRequestDTO },
+      { posting }: { posting: PostingWithShiftsRequestDTO },
```
sherryhli, line 29: The update mutation should eventually also be changed right? We don't need to do it right now but we should make a ticket capturing all the TODOs

backend/services/implementations/postingService.ts
```diff
@@ -182,6 +199,11 @@ class PostingService implements IPostingService {
               return { id: Number(employeeId) };
             }),
           },
+          shifts: {
+            createMany: {
+              data: timeBlocks,
+            },
+          },
```
sherryhli, line 206: I think one way in which this implementation differs from the previous bulk `createShifts` method in `ShiftService` is that we're not performing a conflict check (i.e. calling the `checkConflictingShifts` method to make sure there isn't an existing shift associated with the posting with the exact same startTime and endTime).
https://github.com/uwblueprint/sistering/blob/main/backend/services/implementations/shiftService.ts#L141-L160

A while ago we talked about creating indexes on startTime and endTime instead, should we do that now?

sherryhli (approved):
Looks good to merge to unblock testing but there's a couple of TODOs that should be documented and/or addressed in a separate PR :)

1. Naming of `ShiftDataWithoutPostingId` type
2. `updatePosting` should now likewise include a shifts update
3. Unused mutations and methods (e.g. old bulk `createShifts`) and related types should be deleted
4. Shift creation conflict check

PR #311, opened 2022-05-08
Rickson/content cuts off from role desc screen view on enter press

### LINKED ISSUE #279: Bug

MatoPlus:
## Ticket link

Closes #279 

## Implementation description
* Added missing properties for `draftEditor-content` to avoid cutting cursor view
* Use container for non-growing container width, other wise very long lines could make editor infinitely long, stretching all other components

## Steps to test
1. Go to `http://localhost:3000/admin/posting/create/basic-info`
2. Enter some characters, spam `enter`, cursor should always be visible and within view
3. Enter some really long line, line should wrap instead of growing editor width

## Checklist

### DIFF
--- a/frontend/src/components/common/RichText/RichTextField.tsx
+++ b/frontend/src/components/common/RichText/RichTextField.tsx
@@ -305,7 +305,7 @@ const RichTextField: FunctionComponent<RichTextFieldProps> = (
   };
 
   return (
-    <VStack alignItems="flex-start">
+    <VStack alignItems="flex-start" w="full">
       <HStack h="32px">
         <Select
           ml="12px"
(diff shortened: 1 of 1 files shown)


frontend/src/components/common/RichText/RichTextField.tsx
```diff
@@ -400,25 +401,25 @@ const RichTextField: FunctionComponent<RichTextFieldProps> = (
           ))}
         </ButtonGroup>
       </HStack>
-      <Box
-        w="full"
+      <Container
```
albertlai431, line 404: I think we might wanna leave the `w="full"` here since the width isn't going the full length?

Should be like 
image

instead of 
image

albertlai431 (changes requested):
I think your fix worked for the most part! But when I try doing it with lists (i.e., making a bullet list, typing and entering a bunch), the problem still remains - is it an easy fix to get it working with lists?

frontend/src/components/common/RichText/RichTextField.tsx
albertlai431: Although hm I just saw that there was an issue with growing width... maybe we could set a max width too?

MatoPlus: I've tried that initially as well but the thing is that the "max-width" varies from screen resolutions and we can't guarantee the width always lines up. I'll try digging around a bit though 🤔

LenaNguyen (approved):
Works for me! I couldn't figure out how to trigger the issue with lists that Albert mentioned. It seems uncommon of a scenario enough that it can be handled later

albertlai431:
> Works for me! I couldn't figure out how to trigger the issue with lists that Albert mentioned. It seems uncommon of a scenario enough that it can be handled later

Sounds good, I'll approve!

albertlai431 (approved):

PR #323, opened 2022-05-11
filter by day click

LenaNguyen:
## Steps to test
1. Go to /admin/schedule/posting/1
2. Click on a day
3. Check that the shifts filter to that day
There is a bug for some shift times (seems to be the ones closer to 12 like 1am-3am so it may be due to timezone issues) where the day the shift is being displayed on the calendar is +/- 1 day of the actual shift date :') But it should work sometimes so...yah

## Checklist
- [ ] My PR name is descriptive and in imperative tense
- [ ] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [ ] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

### DIFF
--- a/frontend/src/components/admin/ShiftCalendar/MonthlyViewReadOnlyShiftCalendar.tsx
+++ b/frontend/src/components/admin/ShiftCalendar/MonthlyViewReadOnlyShiftCalendar.tsx
@@ -5,12 +5,14 @@ import FullCalendar, {
   EventInput,
 } from "@fullcalendar/react";
 import dayGridPlugin from "@fullcalendar/daygrid";
+import interactionPlugin from "@fullcalendar/interaction";
 import { Box } from "@chakra-ui/react";
 
 import { Event } from "../../../types/CalendarTypes";
 import colors from "../../../theme/colors";
 import "./Calendar.css";
 import { AdminScheduleShiftWithSignupAndVolunteerGraphQLResponseDTO } from "../../../types/api/ShiftTypes";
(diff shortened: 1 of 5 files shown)


lambo-liu (approved):
Clicking on an event within a day doesn't work (the day div itself works). We can fix it in a subsequent PR.

frontend/src/components/admin/ShiftCalendar/MonthlyViewReadOnlyShiftCalendar.tsx
```diff
@@ -78,6 +82,9 @@ const MonthlyViewShiftCalendar = ({
         displayEventEnd
         dayCellClassNames={(day: DayCellContentArg) => applyCellClasses(day)}
         editable
+        dateClick={({ date }) =>
+          onDayClick(new Date(getISOStringDateTime(date)))
```
sherryhli, line 86: Re: the suspected timezone bug

I think it's because the date constructor here is parsing the ISO datetime string in the local timezone while we want to treat all dates as UTC.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse#date_time_string_format

`getISOStringDateTime` returns `YYYY-MM-DDTHH:mm` which does not specify timezone, thus it gets parsed in local time. We should append `:ss+00:00` to specify UTC. Can use the `getUTCDateForDateTimeString` function from our datetime utils.

PR #358, opened 2022-05-17
Logout user when token is expired

### LINKED ISSUE #328: Logout user when token expires

LenaNguyen:
## Ticket link

Closes #328 

## Implementation description
Currently, when you log in and don't log out, after a few hours you will be unable to properly load data in the application and you will be unable to logout. The issues is that calling FirebaseAdmin to refresh the token throws an `TOKEN_EXPIRED` [error code](https://firebase.google.com/docs/reference/rest/auth/#section-refresh-token). This means that the refresh token (stored in the cookies) is expired.

This solution will logout the user when we fail to refresh the token, because it assumes that the refresh token has expired. The only way to get a new one is to login again.

In the future, we would want the refresh token to not expire so frequently, but this requires more research. 

## Steps to test
Test locally:
1. Go to `frontend/src/index.tsx` and replace the `if` condition on line 59 to be `if(true)`
2. Add `throw new Error()` to line 64 to simulate the api call failing. You can also throw the error in the backend
3. Login and refresh the page
4. It should log you out

## Checklist

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -23,6 +23,24 @@ const REFRESH_MUTATION = `
   }
 `;
 
+const LOGOUT = `
+  mutation Index_Logout($userId: ID!) {
+    logout(userId: $userId)
+  }
+`;
+
+const logout = async (userId: string) => {
+  const { data: result } = await axios.post(
+    `${process.env.REACT_APP_BACKEND_URL}/graphql`,
+    { query: LOGOUT, variables: { userId } },
+    { withCredentials: true },
+  );
+
+  if (result?.data?.logout === null) {
+    localStorage.removeItem(AUTHENTICATED_USER_KEY);
+  }
+};
+
 const link = createUploadLink({
   uri: `${process.env.REACT_APP_BACKEND_URL}/graphql`,
   credentials: "include",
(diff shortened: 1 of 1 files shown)


frontend/src/index.tsx
```diff
@@ -44,19 +62,25 @@ const authLink = setContext(async (_, { headers }) => {
...
-      token = accessToken;
+        const accessToken: string = data.data.refresh;
+        setLocalStorageObjProperty(
+          AUTHENTICATED_USER_KEY,
+          "accessToken",
+          accessToken,
+        );
+        token = accessToken;
+      } catch {
+        await logout(
```
sherryhli, line 80: This will always hit the error case in the `logout` function right? Since this code block only executes when both the access token and refresh token are expired, is there a point to running the logout mutation at all? I think it will always fail.

No need to change right now though

sherryhli (approved):
LGTM, thanks for looking into this and finding a workaround! 🚢

frontend/src/index.tsx
LenaNguyen: oh true. I'm calling the logout function so that it revokes the current tokens and then when you sign back in it issues a new token. But maybe we'll always be given a new refresh token when we login?

PR #379, opened 2022-05-30
Auth Navbar when users not logged in

### LINKED ISSUE #219: "Not logged in" navbar
### Summary and Background
Navbar for users who are not logged in.

### Solution (Core Elements)
Create a navbar to be at the top of the screen whenever no one is logged in.

Left: 'Back to Main' button redirects to Sistering's main page (https://sistering.org/). Should confirm this with Aoife.

Right: Profile dropdown that leads to sign in page #214

### Mockups

Other designs: https://www.figma.com/file/tOK6AjkaRZ1NBXo191D64C/Sistering-%2F-User-Authentication-Process?node-id=390%3A1975

### Technical
Create a component called `AuthNavbar.tsx` in `frontend\src\components\auth`. 
* Wrap the Button component of variant "link" in the `Link` component from `react-router-dom`. This Link component should receive prop `to="/"` .
* see `VolunteerNavbar.tsx` for an example

eevanwong:
## Ticket link

Closes #219 

## Implementation description
* Added navbar when user is not logged in (for all pages that can be displayed when user is not logged in)
image

## Steps to test
1. Go to any page that is allowed when not logged in (e.g localhost:3000/login)

## Checklist


frontend/src/App.tsx
```diff
@@ -104,6 +105,7 @@ const App = (): React.ReactElement => {
                 value={dispatchPostingContextUpdate}
               >
                 <Router>
+                  {authenticatedUser ? "" : <AuthNavbar />}
```
MatoPlus, line 108:
```suggestion
                  {authenticatedUser ? undefined : <AuthNavbar />}
```
would this make more sense semantically?

frontend/src/components/auth/AuthNavbar.tsx
```diff
@@ -0,0 +1,46 @@
...
+        >
+          Back to Main
+        </Button>
+        <Menu>
+          <MenuButton
+            textStyle="button-semibold"
+            color="text.gray"
+            _hover={{ color: "teal" }}
+            _active={{ color: "teal" }}
+            mr="20px"
```
MatoPlus, line 33: A follow-up item is to make this the default style for the `MenuButton` component by editing `theme/index.ts` to reduce redundancy a bit. This is optional

frontend/src/components/auth/AuthNavbar.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <Box px="90px" boxShadow="md">
+      <Flex h="80px" alignItems="center" justifyContent="space-between">
+        <Button
+          leftIcon={<ArrowBackIcon w={8} h={8} />}
+          variant="link"
+          onClick={() => window.open("https://sistering.org/", "_blank")}
+        >
+          Back to Main
+        </Button>
```
MatoPlus, line 26: Unless @JoshuaYe said otherwise, I think opening the link in the current page would flow better (rather than open new tab)

Let's try to avoid `window.open` and use Chakra UI's built in link component

```suggestion
        <Link href="https://sistering.org/">
          <Button leftIcon={<ArrowBackIcon w={8} h={8} />} variant="link">
            Back to Main
          </Button>
        </Link>
```

MatoPlus (changes requested):
Looks great! Just couple of follow-ups from me and it's ready to merge : )

frontend/src/App.tsx
MatoPlus, line 108: This is smart but doesn't quite work out: https://github.com/uwblueprint/sistering/pull/381
we have another navbar for unauth users in the account creation flow.

frontend/src/components/auth/AuthNavbar.tsx
eevanwong: I'm a bit uncertain if what I pushed is what you wanted. I initially tried doing adding it for the MenuButton component like so:
image

image

But the styles did not change. It only changed when I implemented the MenuButton as this, and set a `menuButton` variant for the button.

frontend/src/App.tsx
MatoPlus: hmm sure, I think we can assume that for now but honestly, I don' think there are too many public pages anyways as we currently just manually insert the appropriate navbar into each page. For now, we can go with the suggestion you provided for now.

frontend/src/components/auth/AuthNavbar.tsx
MatoPlus: I think your workaround is good! Chakra UI theme can be a bit ambiguous at times.

frontend/src/App.tsx
eevanwong: On second thought that doesnt work as putting it in App.tsx would not function unless the component that is using the useLocation is in the encapsulated by the Router component. 

Ig this can be an improvement for future stuff. We can put all navbars in a single component and make it dynamic based on routes. I'll go with just manually insertting it into the public components.

MatoPlus (approved):
LGTM

PR #393, opened 2022-06-04
turned environment vars to public constants for hotjar and google ana…

### LINKED ISSUE #386: Fix Google Analytics and Hotjar
# Context
Product @ Blueprint doesn't really use data to make decisions because of the shortness of projects and handoffs. Thus, I thought it'd be cool to break that trend starting with Sistering's soft launch! By integrating Google Analytics and Hotjar, I'm hoping to get some insights for how we can improve next term. 😁

# Technical notes
- We already integrated Hotjar and GA in #298 however, since Firebase Hosting does not allow storage of environmental variables due to its static hosting nature, we want to update the frontend to use JS constants instead of env vars
- Note that it is okay to expose the `TRACKING_ID`, `HJID`, `HJSV` into constants in `App.tsx`, they are not secrets
- Change instances of `process.env.xyz` to use constants instead
- Once this is done, the secret in the vault needs to be cleaned up (this can be done by one of the PLs)

eevanwong:
…lytics

## Ticket link

Closes #386 

## Implementation description
*Changed environment variables to constants, as firebase could not use environment variables which screwed up google analytics

## Steps to test
1. Unsure if thered be any way to test this on a non production environment

## Checklist

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -56,17 +56,19 @@ import PasswordResetSuccessPage from "./components/auth/PasswordResetSuccess";
 import NewAccountPage from "./components/pages/NewAccountPage";
 import AccountCreatedPage from "./components/pages/AccountCreatedPage";
 
-ReactGA.initialize(process.env.TRACKING_ID ?? "");
+// Consts for Hotjar and Google Analytics (this is ok to expose)
+const TRACKING_ID = "G-DF2BP4T8YQ";
+const HJID = "2949419";
+const HSJV = "6";
+
+ReactGA.initialize(TRACKING_ID);
 
 const App = (): React.ReactElement => {
   useEffect(() => {
     ReactGA.pageview(window.location.pathname + window.location.search);
   }, []);
   useEffect(() => {
-    hotjar.initialize(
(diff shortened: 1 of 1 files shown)


frontend/src/App.tsx
```diff
@@ -56,17 +56,19 @@ import PasswordResetSuccessPage from "./components/auth/PasswordResetSuccess";
 import NewAccountPage from "./components/pages/NewAccountPage";
 import AccountCreatedPage from "./components/pages/AccountCreatedPage";
 
-ReactGA.initialize(process.env.TRACKING_ID ?? "");
+// Consts for Hotjar and Google Analytics (this is ok to expose)
+const TRACKING_ID = "G-DF2BP4T8YQ";
+const HJID = "2949419";
+const HSJV = "6";
```
MatoPlus, line 62:
```suggestion
const HJID = 2949419;
const HSJV = 6;
```

If we're just using them as ints, let's just instantiate them as such

MatoPlus (changes requested):
nit

MatoPlus (approved):
LGTM!

PR #462, opened 2022-07-03
Add conditions to render event vs regular postings on postin cards

### LINKED ISSUE #254: Conditionally render shift times in posting card based on event vs opportunity
### Acceptance Criteria
Shift times are only displayed for events, "See posting details" is displayed for regular opportunities.

### Technical Notes
* Add additional `type` prop to `PostingCard` component (exactly like [`EmptyPostingCard`](https://github.com/uwblueprint/sistering/blob/main/frontend/src/components/volunteer/EmptyPostingCard.tsx#L5). Though the type could be determined from other existing props, we should pass it in rather than computing in the component so that the logic for determining event vs opportunity is not duplicated
* Update code here: https://github.com/uwblueprint/sistering/blob/main/frontend/src/components/volunteer/PostingCard.tsx#L51-L55
* Update `frontend/src/components/pages/volunteer/posting/VolunteerPostingsPage.tsx` to pass the `type` prop to `PostingCard`

PL Comments:
- To differentiate between events and regular volunteer opportunities, look at the start/end date, event only last 1 day vs opportunities last consecutive days 
- https://github.com/uwblueprint/sistering/blob/main/frontend/src/components/volunteer/EmptyPostingCard.tsx#L5 <- we should move this type to be used everywhere

MatoPlus:
## Ticket link

Closes #254

## Implementation description
- Add new conditions for regular vs event posting cards
- Add logic to render time range for event posting based on all children 

Regular
image

Event
image

NOTE: An event posting is a posting whose start/end date are the same (EDT time)

## Steps to test
1. Add an event and regular posting that is visible on browse posting page
2. Go to `volunteer/postings`
3. Ensure that you can see the event card with specific date and specified time range
4. Ensure that you can see the regular posting card with date range and no specific time range on the card instead saying "See posting details"

* Correctness and cleaniness

## Checklist

### DIFF
--- a/frontend/src/components/volunteer/PostingCard.tsx
+++ b/frontend/src/components/volunteer/PostingCard.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
 import { TimeIcon, CalendarIcon } from "@chakra-ui/icons";
 import {
   Text,
@@ -11,9 +11,15 @@ import {
   Tag,
 } from "@chakra-ui/react";
 
+import moment from "moment";
 import RichTextDisplay from "../common/RichText/RichTextDisplay";
-import { formatDateStringYear } from "../../utils/DateTimeUtils";
+import {
+  formatDateStringYear,
+  formatTimeHourMinutes,
+} from "../../utils/DateTimeUtils";
 import { SkillResponseDTO } from "../../types/api/SkillTypes";
+import { PostingRecurrenceType } from "../../types/PostingTypes";
+import { ShiftDTO } from "../../types/api/ShiftTypes";
 
 type PostingCardProps = {
   id: string;
@@ -24,6 +30,8 @@ type PostingCardProps = {
   endDate: string;
   autoClosingDate: string;
   branchName: string;
+  type: PostingRecurrenceType;
+  shifts: ShiftDTO[];
   navigateToDetails: () => void;
   navigateToSubmitAvailabilities?: () => void;
   navigateToAdminSchedule?: () => void;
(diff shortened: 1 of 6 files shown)


frontend/src/components/volunteer/PostingCard.tsx
```diff
@@ -38,10 +46,32 @@ const PostingCard = ({
...
+        .map((shift) => moment(new Date(shift.endTime)))
+        .sort((a, b) => a.diff(b));
+      if (sortedStartTimes.length > 0) {
+        setEventFirstTime(sortedStartTimes[0].toDate());
+      }
+      if (sortedEndTimes.length > 0) {
+        setEventLatestTime(sortedEndTimes[sortedEndTimes.length - 1].toDate());
+      }
+    }
+  }, [shifts, type]);
```
MatoPlus, line 73: This is currently missing a check to ensure that all shifts are on the same date as the posting date but I'm not sure if we should be that strict when it comes to this kind of verification. Thoughts?

twilkhoo: I think it's kind of unnecessary- if I understand correctly, shifts for a single "event" must have the same posting date, and if they don't then we need to fix that in the shift creation logic, otherwise it can't be considered an "event" right?

twilkhoo (approved):
Just check my comment on your suggestion, other than that LGTM, the site works as intended

frontend/src/components/volunteer/PostingCard.tsx
MatoPlus: Good point. I also think it's "unnecessary" if we want to depend solely on the backend for consistent shifts but I'm not sure if it's appropriate to leave this out as a consideration on the client-side. I brought this up since I like to directly change stuff in the db a lot. That said, I don't think there's a good solution on the client-side for this case either way. Will resolve for now.

PR #467, opened 2022-07-05
Updated InviteUser Model

eevanwong:
## Implementation description
* Types in user invite were wrong, changed to proper typing, made migration.


MatoPlus (approved):

backend/prisma/schema.prisma
```diff
@@ -166,8 +166,8 @@ model Employee {
 
 model UserInvite {
   pid   String @id @default(uuid())
```
MatoPlus, line 168: TODO: this field should be named `uuid`

Issue #426, opened 2022-06-18
integrate and hook up query data with admin homepage components


MatoPlus:
Use query data to display in homepage. We want to query all the postings and branches.

## Mockup

![image.png](https://images.zenhubusercontent.com/61e48df22f0c8deac5a38226/e606369e-d1d4-4b1c-98e0-c65dc49c85b8)

Figma: https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Sistering-%2F-Admin-Experience?node-id=1421%3A39619

## Acceptance Criteria:
- For this ticket, each of the sections under the tabs should be filled out (unscheduled, scheduled, past, draft)
- The options for the branches dropdown should be populated accordingly

## Technical:
- Use existing admin posting card created in #344 and #343 create them dynamically for each tab

PR #487, opened 2022-07-11
Integrate query data with admin homepage

### LINKED ISSUE #422: call query for posting cards for admin homepage
We want to make apollo client calls to get posting cards in the admin homepage at `/admin/`

Currently the entire admin homepage is hardcoded, we want to make the effort to query real data from our backend.

## Mockup

Figma: https://www.figma.com/file/TWJTNWjK75A6VgSisySTLe/Sistering-%2F-Admin-Experience?node-id=1421%3A39619

## Technical:
- Create a new mutation call in `/admin` to get all the posting of all status
- Create any new graphql request types as required
- You should be able to use the existing `postings` query to fetch all the needed data for posting cards

briantu:
## Ticket link

Closes #422, #426

## Implementation description
* Added query data for posting cards in admin homepage
* Added query data for branches in admin homepage
* Postings should be sorted under each status tab (unscheduled, scheduled, past, draft)

## Steps to test
1. Go to `http://localhost:3000/admin` and check that postings are properly shown under each tab
2. Check branches are shown in the branch dropdown

* Code correctness

## Checklist

### DIFF
--- a/frontend/src/components/pages/admin/AdminHomepage.tsx
+++ b/frontend/src/components/pages/admin/AdminHomepage.tsx
@@ -1,14 +1,82 @@
-import React, { useContext } from "react";
-import { Flex } from "@chakra-ui/react";
+import React, { useContext, useState } from "react";
+import { Flex, Box, SimpleGrid } from "@chakra-ui/react";
+import { generatePath, useHistory } from "react-router-dom";
+import { gql, useQuery } from "@apollo/client";
+import AuthContext from "../../../contexts/AuthContext";
+import * as Routes from "../../../constants/Routes";
 
 import Navbar from "../../common/Navbar";
 import AdminHomepageHeader from "../../admin/AdminHomepageHeader";
+import AdminPostingCard from "../../admin/AdminPostingCard";
 import { AdminNavbarTabs, AdminPages } from "../../../constants/Tabs";
-import AuthContext from "../../../contexts/AuthContext";
 import { Role } from "../../../types/AuthTypes";
+import { PostingResponseDTO } from "../../../types/api/PostingTypes";
+import { isPast } from "../../../utils/DateTimeUtils";
 
-const AdminHomepageComponent = (): React.ReactElement => {
+type Posting = Omit<PostingResponseDTO, "employees" | "type">;
+
+enum PostingStatus {
+  DRAFT = "DRAFT",
+  SCHEDULED = "SCHEDULED",
+  UNSCHEDULED = "UNSCHEDULED",
+  PAST = "PAST",
+}
+
+const POSTINGS = gql`
+  query AdminHomepage_postings {
+    postings {
+      id
+      branch {
+        id
+        name
+      }
+      skills {
+        id
+        name
+      }
+      shifts {
+        id
+      }
+      title
+      description
+      startDate
+      endDate
+      numVolunteers
+      autoClosingDate
+      status
+    }
+  }
+`;
+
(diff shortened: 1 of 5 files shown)


frontend/src/components/pages/admin/AdminHomepage.tsx
```diff
@@ -1,14 +1,112 @@
...
+        }
+        if (postingStatus === PostingStatus.PAST) {
+          sortedPostings[2].push(posting);
+        } else {
+          sortedPostings[3].push(posting);
+        }
+      });
+      setPostingsByStatus(sortedPostings);
+    }
+  }, [postings]);
```
MatoPlus, line 109: pointing this lint error out
`React Hook useEffect has a missing dependency: 'getPostingStatus'. Either include it or remove the dependency array`

MatoPlus (commented):
The use of `postingStatusNums` seems a bit questionable as a pattern but it looks fine. Will leave it to @neelismail01 to stamp

frontend/src/components/pages/admin/AdminHomepage.tsx
briantu: Yeah, I'm not sure how I can fix that. I only want that useEffect to run when `postings` is modified. I tried both adding the other dependencies and removing the array, but that caused the postings to be infinitely added to `postingsByStatus`.

MatoPlus: I think the reason why you're getting an infinite loop is because when you add `getPostingStatus` is because the function is "reevaluated" each time the page is loaded, this hook will also reload the page each time the dependency change so that leads to an infinite loop. You can avoid this by putting getPostingStatus in a util file and importing it so it doesn't get re-evaluated each time it goes to the body of the component

briantu: Ah okay that makes sense. I ended up just moving the function outside of the component, which works as well. Do you still want the util file?

MatoPlus: nah that also works!

frontend/src/components/pages/admin/AdminHomepage.tsx
```diff
@@ -1,14 +1,111 @@
...
+    }
+  }
+`;
 
-const AdminHomepageComponent = (): React.ReactElement => {
+const getPostingStatus = (posting: Posting): PostingStatus => {
+  if (posting.status === "DRAFT") {
+    return PostingStatus.DRAFT;
+  }
+  if (isPast(posting.autoClosingDate)) {
```
MatoPlus, line 55: this should technically be `isPast(posting.endDate)` as auto closing date is the date in which signup ends, not when the event ends. Can followup in later.

frontend/src/components/pages/admin/AdminHomepage.tsx
```diff
@@ -1,14 +1,111 @@
...
+    return PostingStatus.PAST;
+  }
+  if (posting.shifts.length === 0) {
```
MatoPlus, line 58: This isn't right either but I can rework this later when I work on the home admin page

MatoPlus: The logic should be, if any of the signup is published, then the posting is scheduled

MatoPlus (approved):
I think this is good to merge! I'll touch up the homepage in my tickets on the homepage flow : )

Issue #484, opened 2022-07-10
create VolunteerEmployee user resolvers should not be role protected


MatoPlus:
the two resolvers to create volunteer and employee users should be public facing

- Instead of role protected, these resolvers should have accept a invite uuid and we want to ensure user email and role of request matches before proceeding to create account
- We want to change the create volunteer/employee resolver request type to require a user invite uuid
- As a check, we want to unique query the user invite with the uuid and the current user's email to ensure they have a valid invite before proceeding with the user creation 
- Main changes are done in `userResolver` type files changed accordingly

Issue #469, opened 2022-07-05
Call getUserInviteByUUID resolver in account creation page


MatoPlus:
vassume the account creation page is provided with a uuid query param, ie: `/route?uuid=xxxxx`. Use this uuid to call the getUserInviteByUUID resolver to determine which account creation form to show (volunteer vs admin)

Issue #468, opened 2022-07-05
Create getUserInviteByUUID resolver


MatoPlus:
We want a resolver to determine if we have a valid user invite given a uuid (this is the primary id). In our user invite follow, users will be sent a user email with a unique link containing a uuid. When we load the account creation page, we want to get the user invite to
1. ensure we have a valid user invite
2. see which role the user is signing up for
3. if the user invite is older than 2 weeks, delete that record and don't return user invite

Issue #444, opened 2022-06-25
Call sendUserInvite resolver on User Invite User management page


MatoPlus:
When we send a user invite on our user management flow, we should create a new user invite for that invitee's email with the associated role

PR #517, opened 2022-07-20
User Account Creation Flow

### LINKED ISSUE #489: Use uuid query param to call create volunteer/employee resolver
`new-account` route will now accept a query param `uuid`, this query param will be passed until the account creation page form is reached. When submitting the form, we also pass this uuid into the form to ensure the user has access to create a new account.

### Technical
- the createVolunteer and createEmployee resolver is used in the account creation forms, we want to inject the uuid into the variables for form submission

eevanwong:
## Ticket link

Closes #489, #399, #444, #440, #484, #469, #468 (this doesnt complete this task just yet, need to check 2 week time limit).

## Implementation description
* Integrated user invites in `/admin/users`, can now send user invites to emails
* users can click on link in email and go through entire account creation flow (create account page start, newaccountform)
* before user is created after form submission, we check if a userinvite exists (if so, create account, then proceed to account creation successful page), then the userinvite is deleted.

## Steps to test
1. Go to `/admin/users` and click on `Add Volunteer` and enter an email
2. Go through account creation process from email (it might be in spam)
3. once completed, try logging in to confirm it works.

Note: If you verify your email (you get another email after you create an account to verify your email with firebase), then you it will prevent you from using that same email to create a new account on sistering to test, so ignore the verification email. 

## Checklist


eevanwong:
@MatoPlus I realize that for #468 we're supposed to check the 2 week limit, but in the schema itself there is no type in there. I believe need to update the schema myself?

MatoPlus:
> @MatoPlus I realize that for #468 we're supposed to check the 2 week limit, but in the schema itself there is no type in there. I believe need to update the schema myself?

What type are you referring to? For the 2 week limit, I was thinking we can just compare the create date to today and if that is older than 2 weeks, we delete it.

backend/graphql/resolvers/userResolvers.ts
```diff
@@ -121,12 +128,12 @@ const userResolvers = {
 
       if (role === "VOLUNTEER") {
         htmlBody = volunteerAccountCreationInviteTemplate(
-          `https://sistering-dev.web.app/create-account?token=${results.uuid}`,
+          `http://localhost:3000/create-account?token=${results.uuid}`,
```
MatoPlus, line 131: Let's add a env var to reference this

frontend/src/components/pages/NewAccountPage.tsx
```diff
@@ -50,6 +98,16 @@ const NewAccountPage = (): React.ReactElement => {
...
+
+    if (!createEmployeeError) {
+      await deleteUserInvite({
+        variables: {
+          email: userInvite?.email,
+        },
+      });
+    }
+
+    history.push("/account-created");
```
MatoPlus, line 110: Would this make more sense?
```suggestion
    if (!createEmployeeError) {
      await deleteUserInvite({
        variables: {
          email: userInvite?.email,
        },
      });
    
      history.push("/account-created");
    }
```

frontend/src/components/pages/NewAccountPage.tsx
```diff
@@ -76,7 +144,7 @@ const NewAccountPage = (): React.ReactElement => {
         <AccountForm
           mode={AccountFormMode.CREATE}
           isAdmin={isAdmin}
-          email="testemail123@domain.com" // TODO: Replace with firebase email
+          email={userInvite?.email} // TODO: Replace with firebase email
```
MatoPlus, line 147: nice! I think you can remove the TODO comment now?

frontend/src/components/user/AccountForm.tsx
```diff
@@ -213,6 +220,17 @@ const AccountForm = ({
...
+  const formatDate = (date: Date) => {
+    let month = `${date.getMonth() + 1}`;
+    let day = `${date.getDate()}`;
+    const year = date.getFullYear();
+
+    if (month.length < 2) month = `0${month}`;
+    if (day.length < 2) day = `0${day}`;
+
+    return [year, month, day].join("-");
+  };
```
MatoPlus, line 232: Can you re-implement this method using moment.js and put it inn `DateTimeUtils`?

MatoPlus (commented):
Some nits and comments. Will try to test this later tonight.

MatoPlus (changes requested):
left some requests

frontend/src/components/user/AccountForm.tsx
eevanwong: Would this be fine? 
```
hireDate: moment(new Date()).format("YYYY-MM-DD"),
```

backend/services/implementations/userService.ts
```diff
@@ -638,6 +666,30 @@ class UserService implements IUserService {
...
     let firebaseUser: firebaseAdmin.auth.UserRecord;
 
+    try {
+      const userInvite = await prisma.userInvite.findUnique({
+        where: {
+          uuid: volunteerUser.token,
+        },
+      });
+
+      if (userInvite == null) {
```
AdrianaCeric, line 676: [extreme nitpick] but ??? 
```suggestion
      if (userInvite === null) {
```

frontend/src/components/user/AccountForm.tsx
MatoPlus: looks good to me!

frontend/src/components/admin/AddVolunteerModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const onSubmit = async () => {
+    await createVolunteerInvite({
+      variables: {
+        email,
+      },
+    });
+
+    if (createUserInviteError) {
+      console.log("Error with creating volunteer");
```
MatoPlus, line 53: can we render the error modal instead on error? See other page for examples

frontend/src/components/admin/AddAdminModal.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const onSubmit = async (userEmail: string) => {
+    await createAdminInvite({
+      variables: {
+        email: userEmail,
+      },
+    });
+
+    if (createUserInviteError) {
+      console.log("Error with creating volunteer");
```
MatoPlus, line 53: can we replace this with an error modal?

eevanwong:
As a heads up, I added an SITE_URL as a env variable in the repos main .env file. This wil be http://localhost:3000 in a local environment, and https://sistering-dev.web.app for prod.

@MatoPlus Can you remind me how to push my .env to vault?

MatoPlus (commented):
This looks good to me from a high level overview. I'll dig deeper and test it out later in the work session today and hopefully get this merged EOD

MatoPlus:
A followup is that a super admin should not be able to invite a user that is already a user. I believe this should be changing the `createUserInvite` resolver

frontend/src/components/pages/NewAccountPage.tsx
```diff
@@ -39,6 +63,33 @@ const NewAccountPage = (): React.ReactElement => {
...
+  useQuery<GetUserInviteResponse>(GET_USER_INVITE, {
+    fetchPolicy: "cache-and-network",
+    variables: {
+      uuid: token,
+    },
+    onCompleted: (data) => {
+      setUserInvite(data.getUserInvite);
+      if (userInvite?.role === "EMPLOYEE") {
+        setIsAdmin(true);
+      }
```
MatoPlus, line 86: This part is buggy from testing. Should not directly use state after using set action as it is not synchronous.
```suggestion
      if (data.getUserInvite?.role === "EMPLOYEE") {
        setIsAdmin(true);
      }
```

MatoPlus (changes requested):
Left suggestion

frontend/src/components/pages/NewAccountPage.tsx
MatoPlus: I think after this gets changed, this PR should be good to change.

backend/services/implementations/userService.ts
```diff
@@ -454,11 +455,66 @@ class UserService implements IUserService {
...
+          email,
+        },
+      });
+
+      if (user !== null) {
+        throw new Error(
+          "Create User Invite failed. Reason = Email already exists",
+        );
+      }
+
```
MatoPlus, line 517: This doesn't work because email is not a field under the user model, it's managed internally via firebase auth, which means you need to call the firebase service to confirm if there already is an user with that email. If it's too much trouble, we can just remove this.

MatoPlus (approved):
Thanks! Feel free to merge when you are ready

MatoPlus (approved):
LGTM!

PR #525, opened 2022-07-26
Integrate User Invites Modal with Backend

### LINKED ISSUE #516: Integrate User Invites Modal with Backend
Follow up on #479

- Query all user invites on page and render them into the user invites modal with a new `userInvites` query which is only accessible by admins
  - You will need to create a new resolver for this!
- When clicking on the right x icon button, delete the corresponding user invite via email with `deleteUserInvite` resolver

briantu:
## Ticket link

Closes #516

## Implementation description
* Created query resolver for all user invites
* Integrated user invites query and delete user mutation

## Steps to test
1. Go to `http://localhost:3000/admin/users`
2. Open invitations modal
3. Create some sample invitations by clicking on "+" button 
4. Open the invitations modal
5. Check that the invitations just created are shown
6. Check that deleting an invitation works

* Code correctness

## Checklist

### DIFF
--- a/backend/graphql/resolvers/userResolvers.ts
+++ b/backend/graphql/resolvers/userResolvers.ts
@@ -56,6 +56,9 @@ const userResolvers = {
     ): Promise<UserInviteResponse> => {
       return userService.getUserInvite(uuid);
     },
+    getUserInvites: async (): Promise<Array<UserInviteResponse>> => {
+      return userService.getUserInvites();
+    },
     // VolunteerUsers
     volunteerUserById: async (
       _parent: undefined,
--- a/backend/services/implementations/userService.ts
+++ b/backend/services/implementations/userService.ts
@@ -49,6 +49,10 @@ export const convertToSkillResponseDTO = (
   });
 };
 
+const isInviteExpired = (createdDate: Date): boolean => {
+  return getWeekDiff(createdDate, new Date()) >= 2;
+};
+
 class UserService implements IUserService {
   async getUserById(userId: string): Promise<UserDTO> {
     let user: User | null;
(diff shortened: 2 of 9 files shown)


backend/graphql/resolvers/userResolvers.ts
```diff
@@ -56,6 +56,9 @@ const userResolvers = {
     ): Promise<UserInviteResponse> => {
       return userService.getUserInvite(uuid);
     },
+    getUserInvites: async (): Promise<Array<UserInviteResponse>> => {
+      return userService.getUserInvites();
+    },
```
MatoPlus, line 61: Thanks for adding this!

MatoPlus (changes requested):
Made 2 suggestions. Overall looks good. Will leave to @neelismail01 to deep dive and stamp

neelismail01 (approved):
The code looks good to me. Once you get Rickson's error toast done, feel free to merge.

backend/services/implementations/userService.ts
```diff
@@ -498,6 +503,44 @@ class UserService implements IUserService {
...
+        if (isInviteExpired(invite.createdAt)) {
+          expiredInvitesUuids.push(invite.uuid);
+        } else {
+          userInvites.push(invite);
+        }
+      });
+
+      await prisma.userInvite.deleteMany({
+        where: { uuid: { in: expiredInvitesUuids } },
+      });
```
sherryhli, line 522: Curious if this adds any noticeable latency to `getUserInvites` queries. I wonder if it might be worth creating a scheduled job for clearing out expired invites, with the added benefit of guaranteed cleanup intervals for the user invite table. Then again, I guess we don't expect there to be loads of expired entries to delete anyways?

sherryhli (commented):
Nice to see this new user invite flow! 🔥

backend/services/implementations/userService.ts
MatoPlus: Good point to bring up! We considered this along with sending out reminder emails on a weekly basis via some cronjob. I think @neelismail01 originally dug into it but there was too many unknowns and we ended removing it from final launch scope.

At the moment, we delete any user invites that are expired when we try to query them. ie: before returning query results, if they are expired (past 2 weeks), we delete them instead.

sherryhli: Got it, makes sense! I think it should work well in practice

MatoPlus (approved):
Looks good to me. We might wanna follow with some kinda empty state for these tables @a-twh

PR #526, opened 2022-07-26
added hotfix to catch error from firebaseAdmin.getUserByEmail

eevanwong:
## Steps to test
1. Create UserInvite and it shouldnt return an error if its an email that hasnt been used before
2. Create a userinvite with one email, then create another with the same email, should return an err
3. Create a userinvite, go through flow, create user, then create a userinvite with the same email, should return an err

## Checklist

### DIFF
--- a/backend/services/implementations/userService.ts
+++ b/backend/services/implementations/userService.ts
@@ -502,9 +502,19 @@ class UserService implements IUserService {
     email: string,
     role: Role,
   ): Promise<UserInviteResponse> {
+    let errMsg;
     try {
       try {
-        await firebaseAdmin.auth().getUserByEmail(email);
+        const firebaseUser = await firebaseAdmin.auth().getUserByEmail(email);
+        const user = await prisma.user.findUnique({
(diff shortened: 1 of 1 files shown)


backend/services/implementations/userService.ts
```diff
@@ -502,18 +502,30 @@ class UserService implements IUserService {
...
+        if (
+          !(
+            getErrorMessage(error) ===
+            "There is no user record corresponding to the provided identifier."
+          )
+        ) {
+          errMsg = `Error occurred that was not from firebase regarding not finding email: ${getErrorMessage(
+            error,
+          )}`;
+        }
```
MatoPlus, line 528: I don't think this is what we want actually. Now we can send user invites to existing users again.

MatoPlus (changes requested):
left request

MatoPlus (commented):
I shouldn't be able to create a user invite for sistering+admin@uwblueprint.org, but I can.

PR #531, opened 2022-07-26
Fix user invite send check + better error toast

MatoPlus:
Same intent as https://github.com/uwblueprint/sistering/pull/526

https://user-images.githubusercontent.com/44826218/181126092-c0d902c0-1f79-4fa6-919b-7b9004db97ce.mp4


eevanwong (approved):
LGTM, tested functionality works. I was trying to think about how to do it without all the nested trys and I completely forgot about callbacks. Nice work!

PR #546, opened 2022-08-02
Remove backend logout call on refresh failure and reload window

### LINKED ISSUE #513: Actually use refresh access token via refresh token

AdrianaCeric:
## Ticket link

Closes #513 

## Implementation description
* Added `UseEffect` in `App.tsx` to refresh pages automatically if Access Token expires

## Steps to test
1. Go to Network tab on the Console Inspect
2. he first 200 GraphQL response and on `Headers` you should see a refresh token. Copy that.
3. Refresh the page
4.  Navigate to at 200 GraphQL response again and the refresh token should be the same as the access token.

## Checklist

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -12,6 +12,7 @@ import { ChakraProvider } from "@chakra-ui/react";
 
 import { hotjar } from "react-hotjar";
 import ReactGA from "react-ga";
+import { gql, useMutation } from "@apollo/client";
 import Login from "./components/auth/Login";
 import Signup from "./components/auth/Signup";
 import ResetPassword from "./components/auth/ResetPassword";
@@ -57,22 +58,22 @@ import AdminUserManagementPage from "./components/pages/admin/user/AdminUserMana
 import CreatePostingPage from "./components/pages/admin/posting/CreatePostingPage";
 import EditAccountPage from "./components/pages/EditAccountPage";
 import EditPostingPage from "./components/pages/admin/posting/EditPostingPage";
+import authAPIClient from "./APIClients/AuthAPIClient";
 
 // Consts for Hotjar and Google Analytics (this is ok to expose)
 const TRACKING_ID = "G-DF2BP4T8YQ";
 const HJID = 2949419;
 const HSJV = 6;
 
+const REFRESH = gql`
+  mutation Refresh {
+    refresh
+  }
+`;
+
 ReactGA.initialize(TRACKING_ID);
 
 const App = (): React.ReactElement => {
-  useEffect(() => {
-    ReactGA.pageview(window.location.pathname + window.location.search);
-  }, []);
-  useEffect(() => {
-    hotjar.initialize(HJID, HSJV);
-  }, []);
-
   const currentUser: AuthenticatedUser = getLocalStorageObj<AuthenticatedUser>(
     AUTHENTICATED_USER_KEY,
   );
@@ -93,6 +94,24 @@ const App = (): React.ReactElement => {
     postingContextReducer,
     DEFAULT_POSTING_CONTEXT,
   );
+
+  const [refresh] = useMutation<{ refresh: string }>(REFRESH);
+
+  useEffect(() => {
+    ReactGA.pageview(window.location.pathname + window.location.search);
+  }, []);
+
+  useEffect(() => {
+    hotjar.initialize(HJID, HSJV);
+  }, []);
+
+  useEffect(() => {
+    authAPIClient.refresh(refresh).then((response) => {
+      if (!response.valueOf()) {
+        setAuthenticatedUser(null);
+      }
+    });
+  }, [refresh]);
 
   return (
     <ChakraProvider theme={customTheme}>
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -12,7 +12,6 @@ import { ChakraProvider } from "@chakra-ui/react";
 
 import { hotjar } from "react-hotjar";
 import ReactGA from "react-ga";
-import { gql, useMutation } from "@apollo/client";
 import Login from "./components/auth/Login";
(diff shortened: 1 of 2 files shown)


MatoPlus (commented):
I haven't dealt with refresh tokens all too much but I think this makes sense. @sherryhli @neelismail01 any thoughts?

frontend/src/App.tsx
```diff
@@ -94,6 +95,24 @@ const App = (): React.ReactElement => {
...
+    hotjar.initialize(HJID, HSJV);
+  }, []);
+
+  useEffect(() => {
+    authAPIClient.refresh(refresh).then((response) => {
+      if (!response.valueOf()) {
+        setAuthenticatedUser(null);
+      }
+    });
+  }, [refresh]);
```
MatoPlus, line 114: I'm looking at https://www.bezkoder.com/react-refresh-token/ and it seems like we can reduce overhead by looking at response error type and only calling refresh when we get a 401 code or token expired message. Looks like we might have to use some kinda apollo interceptor?

MatoPlus: Or maybe https://stackoverflow.com/a/62872754

I think this what we have is a good start.

sherryhli: Yeah, I think there is existing logic for this in `frontend/src/index.tsx`, provided that I understood the intent of this PR correctly.

@MatoPlus @AdrianaCeric would you mind elaborating on the desired behaviour? Looks we're doing a refresh when the `App` component re-renders, in which case I think the existing code in `index.tsx` might be a more robust way to handle refreshes 🙂

frontend/src/App.tsx
```diff
@@ -94,6 +95,24 @@ const App = (): React.ReactElement => {
...
+  useEffect(() => {
+    hotjar.initialize(HJID, HSJV);
+  }, []);
+
+  useEffect(() => {
+    authAPIClient.refresh(refresh).then((response) => {
+      if (!response.valueOf()) {
+        setAuthenticatedUser(null);
+      }
+    });
```
sherryhli, line 113: Does `await` syntax work here? I believe we favour using `await` over `then` in our codebase

MatoPlus: I think we had some problems with using await in a useEffect hook.

frontend/src/App.tsx
MatoPlus: Good callout, I think that makes sense. I remember that there are still times when the "logout" portion doesn't work. IMO, this logic shouldn't be dependent on the backend, (as it could be down for example), we just want to set the auth user to null manually via `setAuthenticatedUser(null);`, it might make sense to just add this logic at the very end after trying to log out.

@AdrianaCeric could we update the catch block of to do `setAuthenticatedUser(null);` after we attempt to log out at https://github.com/uwblueprint/sistering/blob/main/frontend/src/index.tsx#L32-L42 ?

sherryhli: Hmm, one issue is that we don't have access to the auth context in index.tsx, but I agree that we should be setting user to null there. Let's think on this

sherryhli: Ok, I think one way to achieve this is to force a reload with `window.location.reload()` right after removing the user from local storage in index.tsx. This will sync up the auth context as we initialize it using local storage state in App.tsx.

AdrianaCeric: @MatoPlus what do you mean by update the catch block?

frontend/src/App.tsx
```diff
@@ -94,6 +95,16 @@ const App = (): React.ReactElement => {
     DEFAULT_POSTING_CONTEXT,
   );
 
+  const [refresh] = useMutation<{ refresh: string }>(REFRESH);
```
sherryhli, line 98: Looks like all changes in App.tsx can be reverted, no longer being used :)

sherryhli (commented):
index.tsx looks good to me, just a bit of cleanup in App.tsx left

frontend/src/index.tsx
```diff
@@ -22,23 +22,15 @@ const REFRESH_MUTATION = `
     refresh
   }
 `;
-
 const LOGOUT = `
   mutation Index_Logout($userId: ID!) {
     logout(userId: $userId)
   }
 `;
```
sherryhli, line 29: This can be deleted as it's not longer used

frontend/src/index.tsx
```diff
@@ -22,23 +22,15 @@ const REFRESH_MUTATION = `
...
 
 const logout = async (userId: string) => {
```
sherryhli, line 31:
```suggestion
const logout = async () => {
```

frontend/src/App.tsx
```diff
@@ -12,6 +12,7 @@ import { ChakraProvider } from "@chakra-ui/react";
 
 import { hotjar } from "react-hotjar";
 import ReactGA from "react-ga";
+import { gql, useMutation } from "@apollo/client";
```
sherryhli, line 15: Please remove all unused imports

frontend/src/App.tsx
```diff
@@ -94,6 +89,14 @@ const App = (): React.ReactElement => {
...
   );
 
+  useEffect(() => {
+    ReactGA.pageview(window.location.pathname + window.location.search);
+  }, []);
+
+  useEffect(() => {
+    hotjar.initialize(HJID, HSJV);
+  }, []);
+
```
sherryhli, line 99: If there's no particular reason for moving these down here we can revert the change so as to keep the history clean :)

sherryhli (commented):
LGTM, just some nit cleanup tasks left! :)

sherryhli (approved):
LGTM! I suggest renaming this PR as the purpose of the task was altered, maybe something along the lines of "Remove backend logout call on refresh failure and reload window"?

PR #579, opened 2022-08-12
Clear posting context before visiting new create or edit posting pages

### LINKED ISSUE #564: Create Posting context does not clear
Role: Super Admin

Route: create-posting

Description: posting details context save on subsequent create posting actions. They should be cleared after posting has been created the first time.

Screenshot/Video:

sherryhli:
## Ticket link

Closes #564

## Implementation description
* New `ADMIN_POSTING_RESET` reducer action for resetting posting context to default state
* Dispatches this action before beginning a new create or edit posting flow (i.e. before navigating to the create or edit pages from the admin home page)

Note: This addresses a more general case of the issue vs just clearing after saving to the backend. The posting context is modified before the save so we would still see leftover fields from the context if beginning a new create/edit after previously abandoning a create/edit part way through.

## Steps to test
1. Use the preview deploy link and sign in as an admin
2. Create a new posting from the admin home page
3. Try to create another posting and verify that the form and posting context contain the default fields (can use React dev tools)
image
4. Open a posting in edit mode, verify that the form and context contain the expected fields
6. Try to create another posting and verify that the form and posting context contain the default fields and not fields of the last edited posting

* Any other weirdness around the posting context? In hindsight, maintaining fields in global state via context might have opened up room for a lot of edge cases 😬

## Checklist

### DIFF
--- a/frontend/src/components/admin/AdminHomepageHeader.tsx
+++ b/frontend/src/components/admin/AdminHomepageHeader.tsx
@@ -1,4 +1,4 @@
-import React, { Dispatch, SetStateAction, useState } from "react";
+import React, { Dispatch, SetStateAction, useContext, useState } from "react";
 import {
   VStack,
   HStack,
@@ -17,6 +17,8 @@ import {
 import { AddIcon, SearchIcon, SettingsIcon } from "@chakra-ui/icons";
 import { useHistory } from "react-router-dom";
 import BranchManagerModal from "./BranchManagerModal";
+import { ADMIN_CREATE_POSTING_PAGE } from "../../constants/Routes";
+import PostingContextDispatcherContext from "../../contexts/admin/PostingContextDispatcherContext";
 import { BranchResponseDTO } from "../../types/api/BranchTypes";
 
 type AdminHomepageHeaderProps = {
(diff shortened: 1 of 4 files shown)


MatoPlus (commented):
> Any other weirdness around the posting context? In hindsight, maintaining fields in global state via context might have opened up room for a lot of edge cases 😬

An edge case might be people editing two postings back to back, and then instead of clicking edit button, they just go back to the previous page, since there is no explicit reset action being dispatched, the context might still persist.

This is a bit tricky to handle though. I'm not sure if there's any good way to go about this

MatoPlus (approved):
LGTM, clean work around 💪

sherryhli:
> > Any other weirdness around the posting context? In hindsight, maintaining fields in global state via context might have opened up room for a lot of edge cases 😬
> 
> An edge case might be people editing two postings back to back, and then instead of clicking edit button, they just go back to the previous page, since there is no explicit reset action being dispatched, the context might still persist.
> 
> This is a bit tricky to handle though. I'm not sure if there's any good way to go about this
> 
> Edit: Nvm, this actually seems fine in the preview!

Wasn't quite sure how to replicate the scenario you described, but glad that it didn't turn out to be an issue. You're right in that there could be tricky cases that arise from navigating through the browser back and forward buttons - probably a good idea to do some extensive testing around this.

PR #609, opened 2022-08-18
Update getShiftSignupForUser to only return signups for current postings

### LINKED ISSUE #604: Make sure backend shifts only return shifts that not passed
On the volunteer page, we should not be returning postings that has passed their end dates.

MatoPlus:
## Ticket link

Closes #604 

## Implementation description
* Update getShiftSignupsForUser query to only query for signups for postings that have not ended yet. We do not have any usage of this query that needs expired signups.

## Steps to test
1. Do this locally where you have direct access to change the DB records
2. Login as a volunteer
3. Sign up for 3 signups
4. Go to DB and put signup for PUBLISHED, CANCELLED, and PENDING, one for each. They should appear in volunteer homepage signups tab
5. GO to DB and put posting endate to a date in the past
6. The signups for volunteer should no longer be visible
7. Similarly they should not be visible on schedule review page - volunteer profile page

* 

## Checklist

### DIFF
--- a/backend/services/implementations/shiftSignupService.ts
+++ b/backend/services/implementations/shiftSignupService.ts
@@ -9,6 +9,7 @@ import {
 } from "../../types";
 import logger from "../../utilities/logger";
 import { getErrorMessage } from "../../utilities/errorUtils";
+import { getTodayForTZIgnoredUTC } from "../../utilities/dateUtils";
 
 const prisma = new PrismaClient();
 
@@ -39,6 +40,15 @@ class ShiftSignupService implements IShiftSignupService {
           {
             userId: Number(userId),
           },
+          {
+            shift: {
+              posting: {
+                endDate: {
+                  gte: getTodayForTZIgnoredUTC(),
+                },
+              },
+            },
+          },
           signupStatus
             ? {
                 status: signupStatus,
@@ -47,18 +57,7 @@ class ShiftSignupService implements IShiftSignupService {
         ],
       };
       const shiftSignups = await prisma.signup.findMany({
-        where: {
-          AND: [
-            filter,
-            {
-              shift: {
-                posting: {
-                  endDate: { gt: new Date() },
-                },
-              },
-            },
-          ],
-        },
+        where: filter,
         include: {
           shift: {
             include: {
--- a/backend/services/implementations/shiftSignupService.ts
+++ b/backend/services/implementations/shiftSignupService.ts
@@ -47,7 +47,18 @@ class ShiftSignupService implements IShiftSignupService {
         ],
       };
       const shiftSignups = await prisma.signup.findMany({
-        where: filter,
+        where: {
+          AND: [
+            filter,
+            {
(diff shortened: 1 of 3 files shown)


backend/services/implementations/shiftSignupService.ts
```diff
@@ -47,7 +47,18 @@ class ShiftSignupService implements IShiftSignupService {
...
       };
       const shiftSignups = await prisma.signup.findMany({
-        where: filter,
+        where: {
+          AND: [
+            filter,
+            {
+              shift: {
+                posting: {
+                  endDate: { gt: new Date() },
```
sherryhli, line 56: I suspect there might be an edge case due to how we handle timezones. Running this query between 8pm-12am EDT (or 7pm-12am EST) will exclude postings that end on the current day (Eastern time). Let's verify if this is indeed the case and then fix. Suggested solution (pseudocode):

```suggestion
                  endDate: { gt: moment.isDST() ? new Date() - UTC_EDT_OFFSET : new Date() - UTC_EST_OFFSET },
```

**Context:** Our application generally ignores timezones and stores all dates as UTC and interpret them as Toronto local time[^1]. This approach is meant to simplify handling of daylight savings toggling on and off[^2]. Between the time periods mentioned above, `new Date()` would return a time value that's already in the "next" day according to UTC, which ends up excluding postings that end on the current day according to Eastern time.

Let me know if this doesn't make sense 😅 

[^1]: 2022-08-18 00:00 UTC in the database is interpreted as 2022-08-18 00:00 E**D**T by the application, while 2022-12-18 00:00 UTC is interpreted as 2022-12-18 00:00 E**S**T
[^2]: For example, a shift that is created in the summer with start date of 3pm will continue to start at 3pm after end of daylight savings. It will not automatically change to 2pm.

backend/services/implementations/shiftSignupService.ts
```diff
@@ -47,7 +47,18 @@ class ShiftSignupService implements IShiftSignupService {
         ],
       };
       const shiftSignups = await prisma.signup.findMany({
-        where: filter,
+        where: {
+          AND: [
```
sherryhli, line 51: Is `AND` a binary operator? Could we include the end date condition as part of the `filter` above?

sherryhli (changes requested):

backend/services/implementations/shiftSignupService.ts
MatoPlus: will look into it 👍

MatoPlus: tried to resolve. lmk if solution is sketchy : )

MatoPlus: pre-patch while testing, I also did run into the issue that was described. good catch

sherryhli: Looks good, thanks for addressing!

sherryhli (approved):
🚢

Issue #620, opened 2022-08-20
Filter on Volunteer/Admin homepage is inconsistent with same date edge cases


MatoPlus:
The filter behavior on the homepages seems buggy, especially for cases where the end date is "today".

PR #622, opened 2022-08-20
Fix timezone bug in VolunteerPendingShiftsTableRow

sherryhli:
## Ticket link

N/A - quick fix for this issue:
Capture

## Implementation description
* Interpret date in UTC instead of local time in `formatDateMonthDay` util function used in `VolunteerPendingShiftsTableRow`

## Steps to test
1. Login as a volunteer who has pending shifts
2. Verify that the displayed deadline date of the pending shift's associated posting matches what is returned by the backend
Capture1

* N/A

## Checklist

### DIFF
--- a/frontend/src/utils/DateTimeUtils.ts
+++ b/frontend/src/utils/DateTimeUtils.ts
@@ -60,7 +60,7 @@ export const formatDateStringYearAbbrWeekday = (
  */
 export const formatDateMonthDay = (dateStringInput: string): string => {
   const inputAsDate = new Date(dateStringInput);
-  return moment(inputAsDate).format("dddd, MMM D");
+  return moment(inputAsDate).utc().format("dddd, MMM D");
 };
 
 /**


MatoPlus (approved):
LGTM. Good catch. I think we still have a few timezone bugs in our app, especially with edge cases for same-day comparisons.

MatoPlus (commented):
I think this may have also closed #620, I'll need to test it out.

Edit: Nvm, I don't think it did

sherryhli:
> I think this may have also closed #620, I'll need to test it out.
> 
> Edit: Nvm, I don't think it did but it's fine.

The pending shifts row is currently is the only usage of the `formatDateMonthDay` function so the cause must be something else

sherryhli:
> I think we still have a few timezone bugs in our app, especially with edge cases for same-day comparisons.

Definitely possible, another bug bash could be helpful

PR #584, opened 2022-08-13
Implement Organization-Wide Calendar for Admins & Employees

lambo-liu:
## Ticket link

Closes #535 
Closes #555 

## Implementation description
* Added a new query `shiftsWithSignupsAndVolunteers` (of the shiftService) to return all posting's shifts with signups and volunteers.
* Created the org wide calendar page at `/admin/calendar`. This calendar displays the **scheduled/past** signups and volunteers from all postings across Sistering.
* Some clean-up work
  * Renamed IShiftService.ts to shiftService.ts on the backend to keep interface file naming convention.
  * Re-ordered routes for better organization
* Extended `getUserBranchesByUserId` functionality to also work for employees and admins (this is used in `getPostings`, which is then called in the frontend for employees and admins)

## Steps to test
1. Navigate to [](http://localhost:3000/admin/calendar) as a super admin, should see all scheduled and past postings in any branch.
2. Navigate to the same screen as an admin, should only see scheduled and past postings from your own branch.

## What should reviewers focus on?
* Correctly displaying events

## Checklist
- [x] My PR name is descriptive and in imperative tense
- [x] My commit messages are descriptive and in imperative tense. My commits are atomic and trivial commits are squashed or fixup'd into non-trivial commits
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the PL, as well as other devs who have background knowledge on this PR or who will be building on top of this PR

PR #637, opened 2022-08-21
Only show volunteer posting details if user in branch

### LINKED ISSUE #612: Volunteers should not be able to view details of postings outside branch or drafted postings
As a volunteer if I visit https://sistering-dev.web.app/volunteer/posting/{id} of a posting that is outside of my branch, I should be redirected to the 404 page. Similarly, if I visit a posting that is in a draft state, I should be redirected to 404 page instead. Volunteers should not have view access to the posting in the scenario listed above.

TODO: Create same ticket for employees

briantu:
## Ticket link

Closes #612

## Implementation description
* If volunteer is not in posting branch, then redirect to not found page

## Steps to test
1. Log in as volunteer
2. Try to see posting details for branch current volunteer is not in
3. Check you are redirected to not found page

* Code correctness

## Checklist

### DIFF
--- a/backend/graphql/resolvers/postingResolvers.ts
+++ b/backend/graphql/resolvers/postingResolvers.ts
@@ -21,9 +21,9 @@ const postingResolvers = {
   Query: {
     posting: async (
       _parent: undefined,
-      { id }: { id: string },
+      { id, userId }: { id: string; userId?: string },
     ): Promise<PostingResponseDTO> => {
-      return postingService.getPosting(id);
+      return postingService.getPosting(id, userId);
     },
     postings: async (
       _parent: undefined,
--- a/backend/graphql/resolvers/postingResolvers.ts
+++ b/backend/graphql/resolvers/postingResolvers.ts
@@ -1,3 +1,4 @@
+import { ExpressContext } from "apollo-server-express";
 import PostingService from "../../services/implementations/postingService";
 import IPostingService from "../../services/interfaces/postingService";
 import IShiftService from "../../services/interfaces/shiftService";
@@ -21,9 +22,10 @@ const postingResolvers = {
   Query: {
     posting: async (
       _parent: undefined,
-      { id, userId }: { id: string; userId?: string },
+      { id }: { id: string },
+      context: ExpressContext,
     ): Promise<PostingResponseDTO> => {
-      return postingService.getPosting(id, userId);
+      return postingService.getPosting(id, context);
     },
     postings: async (
       _parent: undefined,
--- a/backend/services/implementations/postingService.ts
+++ b/backend/services/implementations/postingService.ts
@@ -195,7 +195,10 @@ class PostingService implements IPostingService {
     );
   }
 
-  async getPosting(postingId: string): Promise<PostingResponseDTO> {
+  async getPosting(
+    postingId: string,
+    userId?: string,
+  ): Promise<PostingResponseDTO> {
     try {
       const posting = await prisma.posting.findUnique({
         where: {
@@ -215,6 +218,13 @@ class PostingService implements IPostingService {
 
       if (!posting) {
         throw new Error(`postingId ${postingId} not found.`);
+      }
+
+      if (userId !== undefined) {
+        const userBranchIds = await this.getUserBranchesByUserId(userId);
+        if (!userBranchIds.includes(posting.branch.id)) {
+          throw new Error(`User is not part of ${posting.branch.name} branch.`);
+        }
       }
 
       const employeeUsers = await this.convertToEmployeeUserResponseDTO(
--- a/backend/services/implementations/postingService.ts
+++ b/backend/services/implementations/postingService.ts
@@ -9,6 +9,9 @@ import {
   Signup,
   Role,
 } from "@prisma/client";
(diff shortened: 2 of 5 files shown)


briantu:
This fix works, but not sure if it's the best way to do it cuz it's a bit messy. Lmk if this is okay - if it is, then I'll also do it for the volunteer posting availabilities page!

frontend/src/components/pages/volunteer/posting/VolunteerPostingDetails.tsx
```diff
@@ -43,60 +46,120 @@ const POSTING = gql`
...
+    }
+    setLoadingUser(false);
+  }, [userBranches]);
+  useEffect(() => {
+    if (isFirstRun2.current) {
+      isFirstRun2.current = false;
+      return;
+    }
+    setLoadingPosting(false);
+  }, [postingDetails]);
```
MatoPlus, line 113: hmm what are we trying to do here? It seems a bit unintuitive to me.

MatoPlus (commented):
hmm would it make more sense to do this on the server side instead? ie: for the posting query, if the current user is a volunteer, make sure the posting is in the same branch as the volunteer, otherwise, on error we redirect instead of showing a toast @briantu would that be better?

frontend/src/components/pages/volunteer/posting/VolunteerPostingDetails.tsx
briantu: ah yeah that's my way of checking when `userBranches` and `postingDetails` have updated cuz the first time UseEffect runs they are both the initial value so im waiting until the 2nd update which is when they finished loading, and I need to know when they are finished loading otherwise it's a pain handling when they are null and not null and stuff

briantu:
> hmm would it make more sense to do this on the server side instead? ie: for the posting query, if the current user is a volunteer, make sure the posting is in the same branch as the volunteer, otherwise, on error we redirect instead of showing a toast @briantu would that be better?
Wait yeah that sounds like a better solution

MatoPlus (changes requested):
as discussed let's fix the visibility of the posting resolver for volunteers in this PR instead. you can look at the `postings` resolver as an example as I believe that already has filtering.


That was done here: https://github.com/uwblueprint/sistering/pull/584/files#diff-1df777bf9199297d8db41bd9b2be9de9ada51c8ed623879ac6a51a11db1253fdR160-R176

frontend/src/components/pages/volunteer/posting/VolunteerPostingDetails.tsx
```diff
@@ -59,11 +61,10 @@ const VolunteerPostingDetails = (): React.ReactElement => {
     history.push(route);
   };
 
-  return !loading && !postingDetails ? (
+  return (!loading && !postingDetails) || error ? (
```
MatoPlus, line 64:
```suggestion
  return (!loading && !postingDetails) || postingDetailError ? (
```
maybe make this more explicit?

backend/graphql/resolvers/postingResolvers.ts
```diff
@@ -21,9 +21,9 @@ const postingResolvers = {
   Query: {
     posting: async (
       _parent: undefined,
-      { id }: { id: string },
+      { id, userId }: { id: string; userId?: string },
```
MatoPlus, line 24: I wonder if we can use the auth token to get the user id somehow. cuz `userId` field is pretty arbitrary (i can just pass it in for whatever user)

backend/services/implementations/postingService.ts
```diff
@@ -217,6 +220,13 @@ class PostingService implements IPostingService {
         throw new Error(`postingId ${postingId} not found.`);
       }
 
+      if (userId !== undefined) {
+        const userBranchIds = await this.getUserBranchesByUserId(userId);
+        if (!userBranchIds.includes(posting.branch.id)) {
+          throw new Error(`User is not part of ${posting.branch.name} branch.`);
+        }
+      }
```
MatoPlus, line 228: this logic only applies for volunteers right?

backend/graphql/resolvers/postingResolvers.ts
briantu: Yes, we can

backend/services/implementations/postingService.ts
briantu: Will be checking using auth token method now, so logic will explicitly check for volunteers

MatoPlus (approved):
LGTM, thanks for getting this done!

backend/graphql/resolvers/postingResolvers.ts
```diff
@@ -22,8 +23,9 @@ const postingResolvers = {
     posting: async (
       _parent: undefined,
       { id }: { id: string },
+      context: ExpressContext,
```
MatoPlus, line 26: nit:
```suggestion
      { req }: { res: Response },
```
we can only pass in what we need

briantu: I originally tried that, but I got some type error when I did `getAccessToken(req)`, but doing `getAccessToken(context.req)` worked.

PR #638, opened 2022-08-22
Update weekly calendar view to restrict invalid start

### LINKED ISSUE #614: Schedule Shifts Valid Range Should Still Restrict Start range When appropriate
eg: none recurring shifts

For logic, we can enable our start validation guard if the recurring is a none type or if start/end date is in the same week. ie: there is no recurring dates:  `moment(start).isSame(end, "week")`

MatoPlus:
## Ticket link

Closes #614 

## Implementation description
* Enhancing the logic I removed in https://github.com/uwblueprint/sistering/pull/578/files#diff-1419dbfb9a8576e06b1e14356bb91202a213dcc954255f30c3019d8b0a7a9c96L128
* we should only specify valid start for non-recurring schedules or schedules with start/end date in the same week. This basic guard is sufficient although it does not consider more specific cases like 2 week posting for 1 month recurring event (which actually should have a start guard)

## Steps to test
1. As a super admin, login and go to create new posting page
2. You should not be able to select shifts on days before the start date (we still allow it for recurring events lasting over than 1 week to allow uneven schedules) 

* Correctness

## Checklist

### DIFF
--- a/frontend/src/components/admin/ShiftCalendar/WeekViewShiftCalendar.tsx
+++ b/frontend/src/components/admin/ShiftCalendar/WeekViewShiftCalendar.tsx
@@ -34,6 +34,7 @@ type ShiftCalendarProps = {
   initialDate?: string;
   startDate: string;
   endDate: string;
+  isRecurringEvents?: boolean;
 };
 
 const WeekViewShiftCalendar = ({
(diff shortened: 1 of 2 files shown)


AdrianaCeric (approved):
Nice!

briantu (changes requested):
Tested in preview, seems like some days are missing in the calendar when startDate and endDate aren't on the same week for `None` recurrence frequency
e.g. Can't see Sunday or Monday
image

MatoPlus:
> Tested in preview, seems like some days are missing in the calendar when startDate and endDate aren't on the same week for `None` recurrence frequency e.g. Can't see Sunday or Monday image

I don't think there's any actionable item to follow on this edge case within our time frame (eg: full month calednar). As discussed, we can probably ignore it?

PR #639, opened 2022-08-22
Update navbar logout to manually clear auth token and refresh

### LINKED ISSUE #626: Update logout navbar onclick to remove auth token rather than calling backend resolver
Instead of depending on the log out resolver to log out, we should just remove the auth token from our local store like in our apollo middleware refresh logic

MatoPlus:
## Ticket link

Closes #626 

## Implementation description
* Update navbar logout action to manually clear auth token in local storage + refresh instead of hard-blocking on backend call. ie: revoke token when possible, but it's not a priority if it fails

## Steps to test
1. Logout still works for all logged in roles

* Correctness

## Checklist


MatoPlus:
merge since the change seems pretty small

PR #640, opened 2022-08-22
Update better toast msg of delete branch failure toast

### LINKED ISSUE #575: Make it explicit that branches should have no dependencies on posting/shifts before deleting
Possible solution: cannot delete branch, currently in use, posting associated to branch, etc...

Restriction: active posting cannot have target branch to be delete. Past and draft should be fine (delete them along with the branch)

Resolution: If the toast for foreign key constraints occur, we should replace the description of the toast message to become "Branch cannot be deleted until all references are removed explicltly". Otherwise, we might want a toast to say "Are you sure you want do delete branch along with past/draft postings with that branch?"

Easier alternative: Just make the error toast more friendly, no need for implicit posting deletes based on status (very messy). Toast description can be "Cannot delete branch {name} because it is still being used by existing postings, please delete or change as required."

MatoPlus:
## Ticket link

Closes #575 

## Implementation description
- Update toast msg description to be static - usually when a user fails to delete a toast, it is because there are existing postings that is still using the branch, the error tells the user to clear those references before proceeding to delete a branch

image

## Steps to test
1. Log in as super admin, try to delete a branch that is used by a toast, you should see the toast error in the screen shot
2. Try creating a new branch and assign yourself to that branch, you should still be able to delete that branch without any error toast

* Any scenarios where this toast error would not make sense - this solution is a tradeoff for simplicity. I can't really see any other internal error scenarios that would cause this issue atm.

## Checklist

### DIFF
--- a/frontend/src/components/admin/BranchManagerTable.tsx
+++ b/frontend/src/components/admin/BranchManagerTable.tsx
@@ -94,8 +94,8 @@ const BranchManagerTable = ({
         });
       } catch (error: unknown) {
         toast({
-          title: "Cannot delete branch",
-          description: `${error}`,
+          title: `Cannot delete branch: "${selectedBranch.name}"`,
+          description: `Branch is still referenced by existing postings.`,
           status: "error",
           duration: 9000,
           isClosable: true,


frontend/src/components/admin/BranchManagerTable.tsx
```diff
@@ -94,8 +94,9 @@ const BranchManagerTable = ({
         });
       } catch (error: unknown) {
         toast({
-          title: "Cannot delete branch",
-          description: `${error}`,
+          title: `Cannot delete branch: "${selectedBranch.name}"`,
+          description: `Please delete or update references to this branch used
+          by existing postings before proceeding.`,
```
MatoPlus, line 99: branch is still referenced in existing postings

frontend/src/components/admin/BranchManagerTable.tsx
```diff
@@ -94,8 +94,9 @@ const BranchManagerTable = ({
         });
       } catch (error: unknown) {
         toast({
-          title: "Cannot delete branch",
-          description: `${error}`,
+          title: `Cannot delete branch: "${selectedBranch.name}"`,
+          description: `Please delete or update references to this branch used
```
briantu, line 98: Is there a better way to word the description? I think it's a bit confusing. Maybe something like `Branch is still referenced in existing postings`?

briantu (changes requested):
Just wondering if we can make the error description clearer. Otherwise, lgtm!

MatoPlus:
merging small change for launch

PR #663, opened 2022-08-28
Create a SearchSelectorField to replace the current SelectorField for adding skills

### LINKED ISSUE #502: update skills field to search field that creates new skills if not exist
Part of account creation flow. We should change the skill dropdown to a search bar that also creates new skills if they do not exists (trimmed)

- The goal is to make the skills field extensible by end users so our skills become a community-built list. We want this to keep the nature of our platform inclusive.
- we want to call the existing createSkills resolver and refetch skills when we update our list of skills
- The surface area are the account forms (create/edit) along with the posting forms (create/edit)

twilkhoo:
## Ticket link

Closes #502 

## Implementation description
* Added a new component, "SearchSelectorField"
* The component allows users to type and select existing skills, but also allows them to add new skills
* New user-added skills are first stored in a set, from which if they are deleted (volunteer changes their mind), they are removed from the set
* Skills are only added to the db when user submits

## Steps to test
1. Login as any user and edit profile
2. Add skills using the new SearchSelectorField component, which includes selecting existing skills and creating new ones
3. Save changes and ensure the changes persist

* Ensure that changes persist after submission and the component works as expected
* Please pay attention to minor details too: if a user types a skill that already exists, the option to add that skill disappears. The user input is formatted to titlecase (and trimmed). Clicking in the dropdown keeps the dropdown open, but clicking outside closes it. 
* Lmk if there's anything that can be made more intuitive

## Checklist

### DIFF
--- a/frontend/src/components/user/AccountForm.tsx
+++ b/frontend/src/components/user/AccountForm.tsx
@@ -1,5 +1,5 @@
 import { Formik, Form } from "formik";
-import { gql, useQuery } from "@apollo/client";
+import { gql, useQuery, useMutation } from "@apollo/client";
 import {
   Box,
   Button,
@@ -10,6 +10,7 @@ import {
   SimpleGrid,
   Text,
   UnorderedList,
+  useToast,
 } from "@chakra-ui/react";
 import React, { useState } from "react";
 import moment from "moment";
@@ -31,6 +32,7 @@ import {
 } from "../../types/api/LanguageTypes";
 import TextField from "./fields/TextField";
 import SelectorField from "./fields/SelectorField";
+import SearchSelectorField from "./fields/SearchSelectorField";
 
 export enum AccountFormMode {
   CREATE,
@@ -42,6 +44,14 @@ const SKILLS = gql`
     skills {
       id
       name
+    }
+  }
+`;
+
+const CREATE_SKILL = gql`
+  mutation AccountForm_CreateSkill($skill: SkillRequestDTO!) {
+    createSkill(skill: $skill) {
+      id
     }
   }
 `;
@@ -123,6 +133,7 @@ const AccountForm = ({
 
   const queryParams = new URLSearchParams(window.location.search);
   const token = queryParams.get("token");
+  const toast = useToast();
 
   const createInitialValues: CreateAccountFormValues = {
     firstName: "",
@@ -171,66 +182,6 @@ const AccountForm = ({
       setLanguages(data.languages);
     },
   });
-
-  const selectSkill = (
-    skill: string,
-    currentSkills: SkillResponseDTO[],
-    setFieldValue: (field: string, value: SkillResponseDTO[]) => void,
-  ) => {
-    // If the skill is not already selected, add it to the list of skills
-    if (!currentSkills.some((s) => s.id === skill)) {
-      const skillName = skills.find((s) => s.id === skill)?.name || "";
-      setFieldValue("skills", [
-        ...currentSkills,
-        {
-          id: skill,
-          name: skillName,
-        },
-      ]);
-    }
-  };
-
-  const deselectSkill = (
-    skill: string,
-    currentSkills: SkillResponseDTO[],
-    setFieldValue: (field: string, value: SkillResponseDTO[]) => void,
-  ) => {
-    // Remove the skill from the list of skills
-    setFieldValue(
-      "skills",
-      currentSkills.filter((s) => s.id !== skill),
-    );
-  };
-
-  const selectLanguage = (
-    language: string,
-    currentLanguages: LanguageResponseDTO[],
-    setFieldValue: (field: string, value: LanguageResponseDTO[]) => void,
-  ) => {
-    // If the language is not already selected, add it to the list of languages
-    if (!currentLanguages.some((l) => l.id === language)) {
-      const languageName = languages.find((l) => l.id === language)?.name || "";
-      setFieldValue("languages", [
-        ...currentLanguages,
-        {
-          id: language,
-          name: languageName,
-        },
-      ]);
-    }
-  };
-
-  const deselectLanguage = (
-    language: string,
-    currentLanguages: LanguageResponseDTO[],
-    setFieldValue: (field: string, value: LanguageResponseDTO[]) => void,
-  ) => {
-    // Remove the language from the list of languages
-    setFieldValue(
-      "languages",
-      currentLanguages.filter((l) => l.id !== language),
-    );
-  };
 
   const createAccount = (values: CreateAccountFormValues): void => {
     if (onVolunteerCreate && onEmployeeCreate) {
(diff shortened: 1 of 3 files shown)


frontend/src/components/user/AccountForm.tsx
```diff
@@ -172,66 +183,6 @@ const AccountForm = ({
...
-    // Remove the language from the list of languages
-    setFieldValue(
-      "languages",
-      currentLanguages.filter((l) => l.id !== language),
-    );
-  };
-
   const createAccount = (values: CreateAccountFormValues): void => {
     if (onVolunteerCreate && onEmployeeCreate) {
       if (isAdmin) {
```
twilkhoo, line 188: Could someone double-check this part? In `editAccount`, we set `branches: []`, but that effectively deletes the branches that were assigned to the user. I've verified this with prisma studio, seems like a bug?

frontend/src/components/user/AccountForm.tsx
```diff
@@ -308,6 +259,130 @@ const AccountForm = ({
...
+    );
+  };
+
+  const handleSubmit = async (
+    values: CreateAccountFormValues | EditAccountFormValues,
+  ) => {
+    const newSkillsArr: string[] = Array.from(addedSkills);
+    const newSkillsInDB: SkillResponseDTO[] = [];
+
+    for (let i = 0; i < newSkillsArr.length; i += 1) {
```
twilkhoo, line 365: `forEach` doesn't work with `await` as intended, and `for ... of` is restricted.

frontend/src/components/user/AccountForm.tsx
MatoPlus: hmm I need to look into this. I've don't think I've ever encountered this bug on staging though

MatoPlus (commented):
Woah, good job with the custom component. It definitely seemed non-trivial as expected. I'll try to take a closer look sometime tmw.

frontend/src/components/user/AccountForm.tsx
```diff
@@ -385,19 +456,22 @@ const AccountForm = ({
...
                     selectSkill(skill, values.skills, setFieldValue)
                   }
                   onDeselect={(skill) =>
                     deselectSkill(skill, values.skills, setFieldValue)
                   }
+                  onCreateNewOption={(newSkill) => addNewSkill(newSkill)}
+                  onDeleteNewOption={(newSkill) => deleteNewSkill(newSkill)}
                 />
               )}
               <SelectorField
```
MatoPlus, line 477: @twilkhoo it would be really nice to have this for languages as well. Do you think it'll be difficult to include that in this PR as well?

frontend/src/components/user/AccountForm.tsx
```diff
@@ -308,6 +259,130 @@ const AccountForm = ({
...
+
+  const handleSubmit = async (
+    values: CreateAccountFormValues | EditAccountFormValues,
+  ) => {
+    const newSkillsArr: string[] = Array.from(addedSkills);
+    const newSkillsInDB: SkillResponseDTO[] = [];
+
+    for (let i = 0; i < newSkillsArr.length; i += 1) {
+      /* eslint-disable-next-line no-await-in-loop */
+      const newSkill = await handleAddSkillToDB(newSkillsArr[i]);
```
MatoPlus, line 367: Would it make more sense to do `handleAddSkillToDb` when we actually click "Click to add `<skill>`"? I think what we have here is fine since it prevents spam but I guess the trade-off is that our dropdown is less responsive as it is not updated with new skills that is added at the moment.

eg:

Added "Tap-dancing" but it only shows in drop-down for future users.

image

MatoPlus (commented):
Another edge case: 

We can't add any skills that are all caps/acronym. I think this is fine though imo since we don't want to give end users too much freedom to add "unknown" acronyms

image

frontend/src/components/user/AccountForm.tsx
```diff
@@ -385,19 +456,22 @@ const AccountForm = ({
...
               {!isAdmin && (
-                <SelectorField
+                <SearchSelectorField
                   id="skills"
                   label="Skills"
                   values={values.skills}
                   options={skills}
-                  placeholder="Select Skills"
+                  addedValues={addedSkills}
+                  placeholder="Type to select skills or add new ones"
```
MatoPlus, line 465: Would it be more concise to say
```suggestion
                  placeholder="Add Skills"
```

frontend/src/components/user/fields/SearchSelectorField.tsx
```diff
@@ -0,0 +1,246 @@
...
+  InputRightElement,
+  Tag,
+  TagCloseButton,
+  Text,
+  Tooltip,
+} from "@chakra-ui/react";
+import React, { useRef, useState } from "react";
+import TagsPopover from "../../common/TagsPopover";
+
+export type Item = {
```
MatoPlus, line 19: I think we temporarily used the term "setting"(s) which applies to skills, languages, and branches (eg: SettingManager). Maybe we can keep it consistent.

```suggestion
export type Setting = {
```

frontend/src/components/user/fields/SearchSelectorField.tsx
```diff
@@ -0,0 +1,246 @@
...
+  label: string;
+  values: Item[];
+  options: Item[];
+  addedValues: Set<string>;
+  placeholder: string;
+  tooltip?: React.ReactElement;
+  onSelect: (id: string) => void;
+  onDeselect: (id: string) => void;
+  onCreateNewOption: (name: string) => void;
+  onDeleteNewOption: (name: string) => void;
```
MatoPlus, line 35: It seems like this isn't used right?

frontend/src/components/user/AccountForm.tsx
twilkhoo: Yeah I was thinking that too! I'm planning to add the language field changes in a separate PR though after this one gets submitted for organization.

MatoPlus (changes requested):
Made some suggestions, interested in your thoughts. Very impressed with the custom component overall 😤

frontend/src/components/user/AccountForm.tsx
twilkhoo: I considered adding skills to the db upon click at first but went against that for a couple reasons. Firstly we'd have to expose the `deleteSkill` mutation to the user (since users may change their mind and remove skills that they added) which seems kind of risky and second was what you mentioned about avoiding spam. I do like the idea of displaying user-added values in the dropdown and just added that to this PR. Now, when a user adds a custom value, it shows up in the dropdown and in the Selected Skills menu, and removing that value removes both the tag and the value from the dropdown.

frontend/src/components/user/AccountForm.tsx
twilkhoo: Good idea, done.

frontend/src/components/user/fields/SearchSelectorField.tsx
twilkhoo: Oh I see, tbh I would want to keep it as "Item" since this is what we used in `SelectorField.tsx`, so I'd rather be consistent with that.

twilkhoo:
Regarding the acronyms comment, I suppose that is a limitation with converting everything to title case. If we were to avoid converting to title case however, we would likely have lots of case differences in the db (some users may enter things all lowercase, some all uppercase, etc.). Due to this, I think that keeping the title case conversion outweighs the downsides.

frontend/src/components/user/fields/SearchSelectorField.tsx
MatoPlus: Ah gotchu. Yea let's stick with item then.

frontend/src/components/user/AccountForm.tsx
MatoPlus: wait yea, I think I get the bug. Definitely worth fixing since it's big impact. I'll file a ticket for it.

MatoPlus: I think I can quickly take care of this.

MatoPlus: https://github.com/uwblueprint/sistering/pull/665 PR up

MatoPlus (approved):
LGTM, awesome work!

